PR #8, opened 2019-09-13
allow hex and binary numbers in DATA statements

Frank-Buss:


mist64:
Wow! All versions of Microsoft BASIC should have this patch... It would allow calculations in DATA files if read as an integer/float, right?

Frank-Buss:
I tried it, and looks like it doesn't allow calculations. Maybe some other higher level function would be needed. But then it would be slower. It is right now a bit slower, but shouldn't matter much.

mist64:
We're 8x the speed, we can do these things. :)

PR #11, opened 2019-09-15
Use rules in the Makefile

mrmcgibby:


Makefile
```diff
@@ -5,47 +5,60 @@ ifdef PRERELEASE_VERSION
...
+.PHONY : all
+all: rom.bin rom-c64.bin
 
-	ca65 $(ARGS_KERNAL) -g -DPS2 -DCBDOS $(VERSION_DEFINE) -o kernal/kernal.o kernal/kernal.s
+# C64
 
-	ca65 $(ARGS_MONITOR) -DMACHINE_X16=1 -DCPU_65C02=1 monitor/monitor.s -o monitor/monitor.o
+rom-c64.bin : DEFINES=$(VERSION_DEFINE)
+rom-c64.bin : $(ROM_C64_OBJ) rom-c64.txt rom-c64.cfg
+	ld65 -C rom-c64.cfg -o $@ $(ROM_C64_OBJ) -Ln rom-c64.txt
```
greg-king5: `"rom-c64.txt"` is created by this rule; it must not be a prerequisite.

Makefile
```diff
@@ -5,47 +5,60 @@ ifdef PRERELEASE_VERSION
...
-	ca65 -o charset/charset.o charset/charset.s
+%-c64.o : %.s
+	ca65 $(ARGS) -DC64 $(DEFINES) -o $@ $^
 
-	ld65 -C rom.cfg -o rom.bin basic/basic.o kernal/kernal.o monitor/monitor.o cbdos/fat32.o cbdos/util.o cbdos/matcher.o cbdos/sdcard.o cbdos/spi_r_byte.o cbdos/spi_rw_byte.o cbdos/spi_select_device.o cbdos/spi_deselect.o cbdos/main.o keymap/keymap.o charset/charset.o -Ln rom.txt
+# X16
 
+rom.bin : DEFINES=-DPS2 -DCBDOS -DMACHINE_X16=1 -DCPU_65C02=1
+rom.bin : $(ROM_X16_OBJ) rom.cfg rom.txt
+	ld65 -C rom.cfg -o $@ $(ROM_X16_OBJ) -Ln rom.txt
```
greg-king5: `"rom.txt"` is created; it must not be a prerequisite.

Makefile
mrmcgibby: fixed

Makefile
mrmcgibby: fixed

Makefile
greg-king5: We don't want to run `ld65 -o rom-c64.txt`.  Therefore, they should have separate rules; the binary files should be their prerequisites.  Also,
```
c64: rom-c64.bin rom-c64.txt
x16: rom.bin rom.txt
```
are redundant; the prerequisites on each line run the same recipe.  `c64` and `x16` should depend on only the binaries.

mrmcgibby: I think I did what you're asking for, but not quite sure.  Also, do you want me to squash this so there are fewer commits?  I don't use github too often so I'm not sure what the general rule for that sort of thing is.

mrmcgibby: The rom.txt files are generated at the same time as the rom.bin files.  So they use the same rule to get generated.  rom.txt doesn't depend on rom.bin.  If I do as suggested, then `make rom.txt` will fail.

https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html

mrmcgibby: done

PR #26, opened 2019-09-18
Update iso-8859-15.s for º, Ö, ö

mobluse:
I think no characters should have identical bitmaps in 8x8, because it could lead to errors when coding or writing. There are still characters to fix, but this is a start.

### DIFF
--- a/charset/iso-8859-15.s
+++ b/charset/iso-8859-15.s
@@ -1737,12 +1737,12 @@
 .byte %________
 .byte %________
 .byte %________
-; º (based on CP850 A7)
+; º (based on CP850 A7, but one pixel higher)
 .byte %__███___
 .byte %_██_██__
 .byte %_██_██__
+.byte %_██_██__
 .byte %__███___
-.byte %________
 .byte %________
 .byte %________
 .byte %________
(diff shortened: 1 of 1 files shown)


charset/iso-8859-15.s
```diff
@@ -2292,13 +2292,13 @@
 .byte %_██__██_
 .byte %__████__
 .byte %________
-; ö (based on PXLfont and CP850 94)
+; ö (based on CP437 94, but centered)
```
mist64: ü will be inconsistent with ä, ö, ë etc. Why this change?

Screen Shot 2019-09-19 at 13 08 15

mobluse: When I switched Ö to CP437 I thought I should switch ö to CP437 also, but this is perhaps a bad idea, since ä can probably not be lowered. I think the way Ö got higher would not work for the other O type characters. In that case the only way to make them different would be to lower the lower case o type character. I think we should change ö back to what it was, but it might be necessary to lower the other o type characters (except o and ö). I depends on how the upper case O type character looks. I still think it's a good idea that all characters look different.

mobluse: When you say "based on CP850", do you then mean the 8x8 font on https://github.com/emaste/fontstuff/blob/master/syscons-fonts/cp850-8x8.hex

mist64: I agree that all character should look different. Yes, that cp850 file. I had forgotten that the cp437 and the cp850 fonts are actually different styles. Maybe cp437 would have been a better choice for a base for style consistency with the PXLfont characters. I picked cp850 because it had more of the 8859-15 characters I needed.

Issue #7, opened 2019-09-13
Support SD2IEC API in ROM?


Kroc:
If the X16 ROM used the SD2IEC API for its SD-card access, it would make it much easier to get code working for both the C64 and Commander X16. Gregory Naçu put together all-in-one online documentation here: http://www.c64os.com/post/sd2iecdocumentation

mist64:
Great link, thanks.

Since the SD card code has been written from scratch, we need to get the base API working first:
https://www.pagetable.com/?p=1038

mist64:
Except for REL files, the CMDR-DOS now almost has feature parity with sd2iec, including subdirectories, partitions etc.

Close? Open specific bugs for missing features?

PR #35, opened 2019-09-22
Fix #20 nested JSRFAR

### LINKED ISSUE #20: jsrfar can't be nested
Since jsrfar uses the zero page location savbank, nested invocations will break because they overwrite the saved bank.

vii1:
For now, just a test case to start working from.

Currently, the output is:
```
COPYING CODE TO BANK 0...
COPYING CODE TO BANK 1...
COPYING CODE TO BANK 2...
MAIN PROGRAM CALLING BANK 0...
HELLO FROM BANK ZERO
BANK ZERO CALLING BANK TWO...
HELLO FROM BANK TWO
BANK TWO CALLING BANK ONE...
HELLO FROM BANK ONE
WE RETURNED TO BANK TWO.
WE RETURNED TO BANK TWO.
WE RETURNED TO MAIN PROGRAM.
```

### DIFF
--- a/test/test_jsrfar.s
+++ b/test/test_jsrfar.s
@@ -1,0 +1,137 @@
+	.feature labels_without_colons
+
+	.import __BANK0_LOAD__
+	.import __BANK1_LOAD__
+	.import __BANK2_LOAD__
+
+	.code
+	.org $800-1
+	.word $801
+basic_stub
+	.word @1
+	.word 0
+	.byte $9E	; SYS
+	.byte <(((start / 1000) .mod 10) + $30)  ; SYS address
+	.byte <(((start / 100 ) .mod 10) + $30)
+	.byte <(((start / 10  ) .mod 10) + $30)
+	.byte <(((start       ) .mod 10) + $30)
+	.byte 0
+@1	.word 0
+
+DFLTO   = $9A
+CHROUT  = $FFD2
+PTR	= $FB
+BANKSEL	= $9F61
+JSRFAR	= $FF6E
+
+.macro print addr
+	ldx #<addr
+	ldy #>addr
+	jsr strout
+.endmacro
+
+.macro jsrfar addr, bank
+	jsr JSRFAR
+	.word addr
+	.byte bank
+.endmacro
+
+	;.data
+aCopy	.byte "copying code to bank "
+cCpBank	.byte "0...",0
+aMain0	.byte "main program calling bank 0...",0
+aMainRt	.byte "we returned to main program.",0
+
+	.code
+	.proc start
+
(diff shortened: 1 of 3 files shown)


mist64:
Yay, unit tests!

vii1:
Is it OK if I use 65C02 opcodes in the Kernal?

mist64:
Yes.

mist64:
This was fixed in 67ed67467eb4ecb1c5ef308480fd664cbefb883e 

See
https://github.com/commanderx16/x16-rom/commit/67ed67467eb4ecb1c5ef308480fd664cbefb883e#diff-fb5d3a56a144681961f5afbfdbe9ca06R247

@vii1 Can you check with your test?

vii1:
> @vii1 Can you check with your test?

The program flow is correct now.

I added register & stack checks. It seems JSRFAR is not restoring register values correctly (checked also with debugger to be sure it wasn't a bug in the test)

```
COPYING CODE TO BANK 0...
COPYING CODE TO BANK 1...
COPYING CODE TO BANK 2...
MAIN PROGRAM CALLING BANK 0...
HELLO FROM BANK ZERO
BANK ZERO CALLING BANK TWO...
HELLO FROM BANK TWO
BANK TWO CALLING BANK ONE...
HELLO FROM BANK ONE
üREGISTER Y MISMATCH!
üREGISTER X MISMATCH!
üREGISTER A MISMATCH!
WE RETURNED TO BANK TWO.
üREGISTER Y MISMATCH!
üREGISTER X MISMATCH!
üREGISTER A MISMATCH!
WE RETURNED TO BANK ZERO.
üREGISTER Y MISMATCH!
üREGISTER A MISMATCH!
WE RETURNED TO MAIN PROGRAM.
```

test/test_jsrfar.s
```diff
@@ -33,6 +34,10 @@ JSRFAR	= $03D1
 .endmacro
 
 .macro jsrfar addr, bank
+	pha
+	lda #7
+	sta BANKROM
```
greg-king5: The other five bits in that I/O port should be protected.  Eventually, they will be used by something important.  It's best to anticipate that now, instead of needing to fix this file when those bits are assigned.

BruceMcF: Quite. Ifthey are not used for something internally and go onto the User Port, a partial VIA port on the User Port is going to get used for bit banging various serial interfaces, for /selection lines, and for adding more classic parallel port lines to the full VIA port.

    lda BANKROM
    ora #7
    sta BANKROM

mist64:
The original problem has been fixed in the source since.

PR #38, opened 2019-09-24
Allow LOAD with address and add VLOAD

jjgignac:
_There are some issues with this PR that will probably require discussion._

This PR does a few things:
- add a `VLOAD` command to load files directly into video RAM
- extend the `LOAD` command to allow specifying a RAM bank and address
- disable the `VERIFY` command (to make room in BASIC ROM for the other changes)

The new functionality is currently incompatible with the emulator because the emulator intercepts calls to the KERNAL load subroutine for device 1.  I was able to test by temporarily modifying a few of the KERNAL IO subroutines to stream dummy data on device 8.

The `LOAD` command can now be used in either of two forms:
- `LOAD [filename[,device[,relocate]]]` (the original form)
- `LOAD [filename[,device[,bank,address]]]`

In the second form, the RAM bank is switched, then the file is loaded at the specified address.

The `VLOAD` command only supports the second form:
- `VLOAD [filename[,device[,bank,address]]]`

This loads a file into video RAM beginning at the specified bank and address.

There are [differences of opinion on Facebook](https://www.facebook.com/groups/CommanderX16/permalink/524303641654068/) as to whether it's okay to eliminate the `VERIFY` command.  I made sure it was removed in the final commit so it's easy to restore if desired.  I'm willing to modify this PR to expand BASIC ROM into multiple banks, but I thought it would be worth posting as-is first.

### DIFF
--- a/basic/code26.s
+++ b/basic/code26.s
@@ -172,8 +172,7 @@ plsv
 	jsr plsv7       ;get ',fa'
 	ldy #0          ;command 0
 	stx andmsk
-	jsr $ffba
-	jsr paoc20      ;by-pass junk
+	jsr paoc19      ;store then by-pass junk
 	jsr plsv7       ;get ',sa'
 	txa             ;new command
 	tay
@@ -183,6 +182,9 @@ plsv
 ;look for comma followed by byte
 plsv7	jsr paoc30
 	jmp getbyt
+
+;store file parms then maybe end
+paoc19	jsr $ffba
 
 ;skip return if next char is end
 ;
@@ -210,8 +212,7 @@ paoc	lda #0
 	txa
 	ldx #1          ;default device
 	ldy #0          ;default command
-	jsr $ffba       ;store it
-	jsr paoc20      ;skip junk
+	jsr paoc19      ;store then by-pass junk
 	jsr plsv7
 	stx eormsk
 	ldy #0          ;default command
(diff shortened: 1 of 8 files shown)


mist64:
This looks awesome!

There will be no need to remove the `VERIFY` command. In the short term, the RS232 emulator will be replaced with a driver for the VERA UART, saving a few hundred bytes. In the long term, BASIC will occupy multiple banks.

I'll look at the changes more closely later today. I'll also see whether I can remove the RS232 emulator as soon as possible.

mist64:
Oh, and while you are touching the argument parsing of `LOAD`: This would be a good opportunity, to also make sure LOAD can ignore extra garbage like this:

```
LOAD "FILE"       PRG
LOAD "FILE",8     PRG
```

In case the user typed `LOAD` at the beginning of a directory entry.

(I'll change the default device to 8 at some point.)

EgonOlsen71:
I had a go at adding support for this to my compiler. Just to be sure that I got the setup correctly from staring at the code changes:

`LOAD [filename[,device[,bank,address]]]`

-> name and device as usual, bank goes into $9f61, address goes into X/Y and A (the former LOAD/VERIFY flag) has to be set to 0


`VLOAD [filename[,device[,bank,address]]]`

-> name and device as usual, (bank & $f)+1 (the former LOAD/VERIFY flag) goes into A, address goes into X/Y

--> then call load. Is that correct?

jjgignac:
That's all correct except that if the VERIFY command is restored, A will
need to be (bank & $f)+2 for VLOAD.

mist64:
I accidentally merged it, then undid it, now the pull request is closed. :( I think we'll need a new pull request!

jjgignac:
@mist64 No problem.  New PR here: https://github.com/commanderx16/x16-rom/pull/40

PR #40, opened 2019-09-24
Allow LOAD with address and add VLOAD

jjgignac:
_There are some issues with this PR that will probably require discussion._

This PR does a few things:
- add a `VLOAD` command to load files directly into video RAM
- extend the `LOAD` command to allow specifying a RAM bank and address
- disable the `VERIFY` command (to make room in BASIC ROM for the other changes)

The new functionality is currently incompatible with the emulator because the emulator intercepts calls to the KERNAL load subroutine for device 1.  I was able to test by temporarily modifying a few of the KERNAL IO subroutines to stream dummy data on device 8.

The `LOAD` command can now be used in either of two forms:
- `LOAD [filename[,device[,relocate]]]` (the original form)
- `LOAD [filename[,device[,bank,address]]]`

In the second form, the RAM bank is switched, then the file is loaded at the specified address.

The `VLOAD` command only supports the second form:
- `VLOAD [filename[,device[,bank,address]]]`

This loads a file into video RAM beginning at the specified bank and address.

There are [differences of opinion on Facebook](https://www.facebook.com/groups/CommanderX16/permalink/524303641654068/) as to whether it's okay to eliminate the `VERIFY` command.  I made sure it was removed in the final commit so it's easy to restore if desired.  I'm willing to modify this PR to expand BASIC ROM into multiple banks, but I thought it would be worth posting as-is first.


jjgignac:
@mist64 I'll restore the VERIFY command and make it ignore end-of-line garbage, as you requested in #38 .

jjgignac:
@mist64 I pushed a commit that allows trailing garbage after the LOAD command.  This commit actually saves 30 bytes.  😀

Note that there are four lines commented out:
```
;
;eat trailing garbage after address parm
;
;       jsr chrgot
;       beq plsv20
;plsv10 jsr chrget
;       bne plsv10
;
```
These lines are only needed if we want to allow garbage after relocated load commands like this:
```
LOAD "X",8,0,8192 PRG
```
It costs 10 bytes, and is of highly questionable value since typing the parameters will almost always overwrite the garbage.  (My example uses a peculiarly short filename.)

jjgignac:
I was able to make enough size optimizations so the VERIFY command could be reintroduced.

jjgignac:
@mist64 I merged the latest master, but it introduced a build error:
```
ld65: Error: Input file `charset/charset.o' not found
```

mist64:
fixed in 28b879c

mist64:
Sorry I haven't merged this, I have been busy with other things. I will look at this again today, but I'll also release r32, and I'm afraid it won't be able to make the cut of this release, because it'll need some testing before.

mist64:
I manually rebased and merged your changes. :) I squashed the last three into one.

a828630 [BASIC] Optimize the file command parsers
f48be34 [BASIC] Add bank/location parms to load, verify
6ff3c77 [KERNAL] Add support for VLOAD
a5ab887 [BASIC] Add the VLOAD command
09215c7 [BASIC] Allow trailing garbage after LOAD command
8b1f368 [BASIC] Size optimizations

Thanks a lot, awesome work!

PR #44, opened 2019-10-02
README: Add build and some cc65 install instructions

0cjs:
This also notes that a built version of `rom.bin` is included with emulator releases, and makes clear in the README that the ROM version currently must be contemporary with the emulator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-# Commander X16 BASIC/KERNAL/DOS/GEOS ROM
+Commander X16 BASIC/KERNAL/DOS/GEOS ROM
+=======================================
 
 This is the Commander X16 ROM containing BASIC, KERNAL, DOS and GEOS. BASIC and KERNAL are derived from the [Commodore 64 versions](https://github.com/mist64/c64rom). GEOS is derived from the [C64/C128 version](https://github.com/mist64/geos).
 
@@ -9,7 +10,9 @@ This is the Commander X16 ROM containing BASIC, KERNAL, DOS and GEOS. BASIC and 
 	* does not support tape (device 1).
 * GEOS is fully compatible with the C64 version.
 
-## New Features
+
+New Features
+------------
 
 * F-keys:
 	F1: `LIST`
@@ -42,14 +45,18 @@ This is the Commander X16 ROM containing BASIC, KERNAL, DOS and GEOS. BASIC and 
 	* everything else is unimplemented
 * Some new KERNAL APIs (to be documented)
 
-## Big TODOs
+
+Big TODOs
+---------
 
 * DOS needs more features.
 * BASIC needs more features.
 * RS232 and IEC are not working.
 * PS/2 and SD have issues on real hardware.
 
-## ROM Map
+
+ROM Map
+-------
 
 |Bank|Name   |Description                                            |
 |----|-------|-------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -0,0 +1,54 @@
...
+
+If cc65 is not available as a package on your system, you'll need to
+install or build/install it per the instructions below.
+
+Once the prerequisites are available, type `make` to build `rom.bin`.
+To use that with the emulator, copy it to the same directory as
+the `x16emu` binary or use `x16emu -rom .../path/to/rom.bin`.
+
+> __WARNING:__ The emulator will work only with a contemporary version
+> of `rom.bin`; eariler or later versions are likely to fail.
```
greg-king5: Typo: `earlier`

mist64:
Thanks for this. Can we have the build instructions in the README instead? This will make them easier to find.

BUILD.md
0cjs: Thanks! I've force-pushed a new commit with the fixes and will try to remember to use a spell-checker from now on.

0cjs:
@mist64:
> Can we have the build instructions in the README instead? This will make them easier to find.

Sure, no problem. As things get large I tend to prefer separating out the topics not everybody will want to see (build instructions, change log, etc.) and putting just the overview and guide to other files in the README, but of course when things are "starting to get large" is an open question.

However, this brings up another matter. If you've looked at these files in an editor or diff tool, and especially one without syntax highlighting (e.g., `git diff` at the command line), you'll probably have noticed that I wrote `BUILD.md` in a more "formatted" plaintext style designed to be as easily read as plain text as when rendered. (Lines are wrapped rather than arbitrary length, first and second level headings use the alternate format so they stand out more and look like headings instead of comments, links use references rather than being inline, etc.)

This doesn't fit the format of the main README file, so I'd like to know your feelings on handling this.Three options that come to mind are:

1. I fully reformat `README.md` file in this style (as a separate "reformat-only" commit) and then add this information to it.

2. Minimal reformatting of `README.md` just to put the first- and second-level headings into the alternate format (to make scanning the file in text mode easier) and then adding the contents of `BUILD.md` formatted as it currently is (except with its headings pushed down a level).

3. Changing the format of `BUILD.md` to match the current `README` format (just `##` to mark second-level headings, inlining the URLs of references, not wrapping the text) and merging it into `README.md`.

I have a pretty strong preference for the first or second, and in general moving towards "readable plaintext," since I consider the documentation basically to be more code (an extended version of comments) and I edit them simultaneously in my workflow.

That said, it's your project, so what are your thoughts on this?

mist64:
My philosophy is this:

* Readability of the plain text is nice, yes, but will be less and less important as all more and more tools will immediately format MarkDown correctly. And this is the README.md after all, which most people will read on the GitHub page.
* Manual line breaks for readability will lead to unnecessary diffs when changing text, or worse, people avoiding updating documentation because they would have to reformat the text and re-do the line breaks. Also, all reasonable text editors will break lines while viewing anyway, so it's not like anyone will have to scroll horizontally like crazy to be able to read it.

Therefore, I vote for number two. I'm all okay with reformatting headings etc. to make everything prettier, but please no manual line breaks.

0cjs:
That's a good point about the diffs, so yes, number two sounds like a good compromise. I've reworked the branch to implement this.

Issue #47, opened 2019-10-03
LOAD command is broken


qbradq:
Creating a test program and saving it to local file system works:
![](http://i.imgur.com/uPpqBvP.png)

Contents of TEST on local file system:
```
qbradq@qbradq-AB350-Gaming:~/commanderx16$ xxd TEST
00000000: 0108 0f08 0a00 9920 2248 454c 4c4f 2200  ....... "HELLO".
00000010: 1808 1400 8920 3130 0000 00              ..... 10...
```

And here is a new session attempting to load the newly created file:
![](http://i.imgur.com/WL5yp61.png)

Tests were on a fresh pull and build of the emulator and ROM.
ROM version 8b1f368435848b607f3e939c85da295e83441167
Emulator version f3121f06c69283e18fa5421562dbc9f17c413d31

greg-king5:
- They're emulator bugs, not Kernal bugs (the Kernal knows nothing about "local file-systems").  This issue should be moved to that repository.
- `SAVE`, also, has a bug.  It didn't convert that file-name from PetSCII into UTF-8.

PR #51, opened 2019-10-05
Fix a bug in LOAD/VERIFY

jjgignac:
This fixes https://github.com/commanderx16/x16-emulator/issues/162
This fixes https://github.com/commanderx16/x16-rom/issues/47

Due to an incorrect branch, the accumulator wasn't being restored after the call to parse the LOAD parameters.  As a result, the KERNAL was loading the data into video memory instead of normal RAM.

### DIFF
--- a/basic/code26.s
+++ b/basic/code26.s
@@ -82,10 +82,10 @@ cverf	lda #1          ;verify flag
 cload	lda #0          ;load flag
 	pha
 	jsr plsv        ;parse parameters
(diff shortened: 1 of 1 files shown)


rsbohn:
This change does fix the issue, allowing LOAD to work.

qbradq:
Confirmed this patch does allow LOAD to work correctly.

mist64:
Sorry for the delay!

PR #55, opened 2019-10-06
[build] ignore tmp files in .gitignore

GendoIkari:
I know, it's the silliest thing ever, but it was bothering me :P


mist64:
Thanks! It's the little things!

Issue #17, opened 2019-09-16
BASIC: Block IO and Transfer


tomxp411:
2 does not have any commands or functions to enable block transfer of data from storage or within memory on the computer. With the addition of VERA and the banked upper memory space, I believe some block transfer commands are necessary to make BASIC more capable of handling large amounts of data. Because of the overhead involved, PEEK and POKE loops just take too long to populate display memory or load binary data from disk.

I'm proposing some block data transfer commands specifically targeted at moving data in and out of banked memory, as well as reading binary data from disk. This could be used for animation frames, sprites, structured data (ie: map data in RPGs). 

This is one suggested method of dealing with the issue, although there may be other ways of accomplishing the same task. 

I have also suggested two utility routines to make data manipulation easier and to assist using main memory to store assembly routines or raw data. 

I'm happy to provide use cases and ideas separately. This description is already pretty long.

BANK _bankno_
Set the bank number for BASIC PEEK and POKE operations. Any BASIC operations that happen within banked memory will cause _bankno_ to increment internally when the internal pointer used for block moves hits $BFFF. Transfer the byte at $BFFF, then set the pointer to $A000 and increment _bankno_. 

GETBANK()

Returns the current _bankno_. This is NOT the setting of the hardware register, but rather the setting of the BASIC pointer used with block operations. (It's assumed that the OS will need to alter the bank register for internal operations, so the value may need to be stashed and loaded as needed.)

COPY _fromaddr_, _toaddr_, _len_
VCOPY _fromaddr_, _tobank_, _toaddr_, _len_

Block copy data. 
VCOPY copies data to VERA. 
When _fromaddr_ or _toaddr_ is a location inside of the 8K banked memory region, wrap to the next bank.

PEEK$(_address_, _[len]_)

Read from memory (PEEK$). This allows for printing portions of a larger data set. For example, room descriptions in a text adventure or fields in a database. Bank wrapping should occur as needed. The intended use is PRINT PEEK$(.,..) to print text from memory, although it could be useful in LET statements, too, if the data has to be manipulated before printing. 

If _len_ is supplied, the first byte read is the first byte of the string. If _len_ is not supplied, the first byte is the length of the string. Max string length is 255 bytes. 

PEEKI/POKEI
PEEKF/POKEF

Just quick utility methods to transfer floating point and 16-bit integers into memory. This does not set the logical value, but rather simply copies the raw bytes in the same order BASIC uses them. 

GET# _buffer_, _len_

[...]

PR #63, opened 2019-10-14
BASIC: Peek() with optional bank parm

kktos:
Tweaks the peek command to allow an optional bank parm.

PRINT PEEK($200)

PRINT PEEK(4, $C000)

### DIFF
--- a/basic/code10.s
+++ b/basic/code10.s
@@ -1,3 +1,5 @@
+.setcpu "65c02"
+
 ;test pointer to variable to see
 ;if constant is contained in basic.
 ;array variables have zeroes placed
@@ -127,10 +129,16 @@ nesct3
 	tay
 	txa
 	pha
-	jmp fingo
-oknorm	jsr parchk
-	pla
-	tay
+	bra fingo
+;
+; add a special case for PEEK to allow PEEK(addr) and PEEK(bank, addr)
+;
+oknorm	jsr chkopn
+	jsr frmevl
+	ply
+	cpy #fnpeek*2 & $FF
+	beq fingo
+	jsr chkcls
 fingo	lda fundsp-onefun-onefun+256,y
 	sta jmper+1
 	lda fundsp-onefun-onefun+257,y
--- a/basic/code10.s
+++ b/basic/code10.s
@@ -116,29 +116,19 @@ nesct3
 	jsr frmevl
 	jsr chkcom
 	jsr chkstr
-	pla
-	tax
+	plx
 	lda facmo+1
 	pha
 	lda facmo
 	pha
-	txa
-	pha
+	phx
 	jsr getbyt
-	pla
-	tay
-	txa
-	pha
-	bra fingo
-;
-; add a special case for PEEK to allow PEEK(addr) and PEEK(bank, addr)
-;
-oknorm	jsr chkopn
-	jsr frmevl
 	ply
-	cpy #fnpeek*2 & $FF
-	beq fingo
-	jsr chkcls
+	phx
+	bra fingo
+
+oknorm	jsr parchk
+	ply
 fingo	lda fundsp-onefun-onefun+256,y
 	sta jmper+1
 	lda fundsp-onefun-onefun+257,y
--- a/basic/code17.s
+++ b/basic/code17.s
@@ -1,3 +1,5 @@
+.setcpu "65c02"
+
 getbyt	jsr frmnum
 conint	jsr posint
 	ldx facmo
@@ -51,28 +53,44 @@ getadr	lda facsgn
 	sty poker
 	sta poker+1
 	rts
+
+.import fetvec, fetch
 peek	lda poker+1
 	pha
 	lda poker
 	pha
+
 	jsr getadr
-	ldy #0
+
+	jsr chrgot
+	ldx #BANK_KERNAL
+	cmp #','	; check if it's PEEK(bank, addr)
+	bne :+
+
+	ldx poker
+	phx
+	jsr chrget
+	jsr frmadr
+	plx
+
+:	ldy #0
 	lda poker+1
 	cmp #$c0
 	bcs peek1
 	lda (poker),y   ;RAM
 	jmp peek2
 peek1	lda #poker
-.import fetvec, fetch
 	sta fetvec
-	ldx #BANK_KERNAL
 	jsr fetch       ;ROM
-peek2	tay
+peek2	sta poker
+	jsr chkcls
+	ldy poker
 dosgfl	pla
 	sta poker
 	pla
 	sta poker+1
 	jmp sngflt
+
 poke	jsr getnum
 	txa
 	ldy #0
--- a/basic/code17.s
+++ b/basic/code17.s
@@ -38,9 +38,11 @@ st2txt	ldx strng2
 	stx txtptr
 	sty txtptr+1
 valrts	rts
+
 getnum	jsr frmadr
 combyt	jsr chkcom
 	jmp getbyt
+
 frmadr	jsr frmnum
 getadr	lda facsgn
 	bmi gofuc
--- a/basic/code17.s
+++ b/basic/code17.s
@@ -54,48 +54,43 @@ getadr	lda facsgn
 	sta poker+1
 	rts
 
+;
 .import fetvec, fetch
-peek	lda poker+1
+peek
+	lda poker+1
 	pha
 	lda poker
 	pha
-
 	jsr getadr
-
-	jsr chrgot
-	ldx #BANK_KERNAL
-	cmp #','	; check if it's PEEK(bank, addr)
-	bne :+
-
-	ldx poker
-	phx
-	jsr chrget
-	jsr frmadr
-	plx
-
-:	ldy #0
+	ldy #0
 	lda poker+1
-	cmp #$c0
+	cmp #$A0	;main:$00-9F bankRAM:$A0-$BF bankROM:$C0-$FF
 	bcs peek1
-	lda (poker),y   ;RAM
-	jmp peek2
+	lda (poker),y   ;MAIN RAM
+	bra peek2
+
 peek1	lda #poker
 	sta fetvec
-	jsr fetch       ;ROM
-peek2	sta poker
-	jsr chkcls
-	ldy poker
+	ldx membank
+	jsr fetch       ;BANKED RAM & ROM
+
+peek2	tay
 dosgfl	pla
 	sta poker
 	pla
 	sta poker+1
 	jmp sngflt
 
-poke	jsr getnum
+;
+.import stavec, stash
+poke
+	jsr getnum
 	txa
-	ldy #0
-	sta (poker),y
-	rts
+	ldx #poker
+	stx stavec
+	ldx membank
+	jmp stash
+
 fnwait	jsr getnum
 	stx andmsk
 	ldx #0
--- a/basic/code26.s
+++ b/basic/code26.s
@@ -46,7 +46,29 @@ setmsg	=$ff90
 
 plot	=$fff0
 
-csys	jsr frmadr      ;get int. addr
+csave	jsr plsv        ;parse parms
+	bcs nsnerr6
+	jmp snerr6      ;disallow bank/address parms
+nsnerr6	ldx vartab      ;end save addr
+	ldy vartab+1
+	lda #<txttab    ;indirect with start address
+	jsr $ffd8       ;save it
+	bcs erexit
+	rts
+
+;
+.import far_jumper, fjmpaddr, fjmpbank
+csys
+	.byt $ff
+	jsr frmadr      ;get int. addr
+
+	lda poker
+	sta fjmpaddr
+	lda poker+1
+	sta fjmpaddr+1
+	lda membank
+	sta fjmpbank
+
 	lda #>csysrz    ;push return address
 	pha
 	lda #<csysrz
@@ -57,7 +79,9 @@ csys	jsr frmadr      ;get int. addr
 	ldx sxreg
 	ldy syreg
 	plp             ;load 6502 status reg
-	jmp (linnum)    ;go do it
+;	jmp (linnum)    ;go do it
+	jmp far_jumper
+
 csysrz	=*-1            ;return to here
 	php             ;save status reg
 	sta sareg       ;save 6502 regs
@@ -67,15 +91,6 @@ csysrz	=*-1            ;return to here
 	sta spreg
 	rts             ;return to system
 
-csave	jsr plsv        ;parse parms
-	bcs nsnerr6
-	jmp snerr6      ;disallow bank/address parms
-nsnerr6	ldx vartab      ;end save addr
-	ldy vartab+1
-	lda #<txttab    ;indirect with start address
-	jsr $ffd8       ;save it
-	bcs erexit
-	rts
 
 cverf	lda #1          ;verify flag
 	.byt $2c        ;skip two bytes
--- a/basic/code26.s
+++ b/basic/code26.s
@@ -59,7 +59,6 @@ nsnerr6	ldx vartab      ;end save addr
 ;
 .import far_jumper, fjmpaddr, fjmpbank
 csys
-	.byt $ff
 	jsr frmadr      ;get int. addr
 
 	lda #>csysrz    ;push return address
--- a/basic/code26.s
+++ b/basic/code26.s
@@ -62,25 +62,35 @@ csys
 	.byt $ff
 	jsr frmadr      ;get int. addr
 
-	lda poker
-	sta fjmpaddr
-	lda poker+1
-	sta fjmpaddr+1
-	lda membank
-	sta fjmpbank
-
 	lda #>csysrz    ;push return address
 	pha
 	lda #<csysrz
 	pha
 	lda spreg       ;status reg
 	pha
+
+	lda poker+1
+	cmp #$A0
+	bcc shortjump
+
+	sta fjmpaddr+1
+	lda poker
+	sta fjmpaddr
+	lda membank
+	sta fjmpbank
+
 	lda sareg       ;load 6502 regs
 	ldx sxreg
 	ldy syreg
 	plp             ;load 6502 status reg
-;	jmp (linnum)    ;go do it
 	jmp far_jumper
+
+shortjump
+	lda sareg       ;load 6502 regs
+	ldx sxreg
+	ldy syreg
+	plp             ;load 6502 status reg
+	jmp (linnum)    ;go do it
 
 csysrz	=*-1            ;return to here
 	php             ;save status reg
--- a/basic/code5.s
+++ b/basic/code5.s
@@ -79,7 +79,9 @@ onglp1	jsr clnget
 	beq onglop
 	pla
 ongrts	rts
+
 clnget	jsr chrget
+
 linget	ldx #0
 	stx linnum
 	stx linnum+1
@@ -109,6 +111,7 @@ morlin	bcs ongrts
 	inc linnum+1
 nxtlgc	jsr chrget
 	jmp morlin
+
 let	jsr ptrget
 	sta forpnt
 	sty forpnt+1
--- a/basic/code9.s
+++ b/basic/code9.s
@@ -231,11 +231,11 @@ eval4	cmp #fntk
 	jmp isfun
 parchk	jsr chkopn
 	jsr frmevl
-chkcls	lda #41
+chkcls	lda #')'
 	.byt $2c
-chkopn	lda #40
+chkopn	lda #'('
 	.byt $2c
-chkcom	lda #44
+chkcom	lda #','
 synchr	ldy #0
 	cmp (txtptr),y
 	bne snerr
--- a/basic/declare.s
+++ b/basic/declare.s
@@ -12,6 +12,9 @@ strsiz	=3
 .segment "ZPBASIC" : zeropage
 
 .segment "BVARS"
+
+membank	.res 1           ;memory bank used by BANK(),PEEK(),POKE(),SYS()
+
 ;                      C64 location
 ;                         VVV
 charac	.res 1           ;$07 a delimiting character
@@ -230,4 +233,3 @@ sxreg	.res 1           ;$030D .x reg
 syreg	.res 1           ;$030E .y reg
 spreg	.res 1           ;$030F .p reg
 usrpok	.res 3           ;$0310 user function dispatch
-
--- a/basic/macros.inc
+++ b/basic/macros.inc
@@ -1,0 +1,6 @@
+;
+; declare a string with the last char with msb=1 as terminator
+;
+.macro  tokString    str
+	.byte .sprintf (.concat ("%.", .string (.strlen(str) - 1) ,"s%c"), str, $80 + .strat (str, .strlen(str) - 1))
+.endmacro
--- a/basic/token2.s
+++ b/basic/token2.s
@@ -1,3 +1,4 @@
+
 	.byt "TAB",$a8
 tabtk	=$a3
 	.byt "T",$cf
@@ -42,7 +43,7 @@ onefun	=$b4
 	.byt "SI",$ce
 	.byt "TA",$ce
 	.byt "AT",$ce
-	.byt "PEE",$cb
+	tokString "PEEK"
 fnpeek	=$c2
 	.byt "LE",$ce
 	.byt "STR",$a4
@@ -58,22 +59,23 @@ gotk	=$cb
 	.byt 0
 
 ;**************************************
-reslst2	.byt "MO", 'N' + $80
-	.byt "DO", 'S' + $80
-	.byt "OL", 'D' + $80
-	.byt "GEO", 'S' + $80
-	.byt "VPOK", 'E' + $80
-	.byt "VLOA", 'D' + $80
-	.byt "SCREE", 'N' + $80
-	.byt "PSE", 'T' + $80
-	.byt "LIN", 'E' + $80
-	.byt "FRAM", 'E' + $80
-	.byt "REC", 'T' + $80
-	.byt "CHA", 'R' + $80
-	.byt "VPEE", 'K' + $80
+reslst2	tokString "MON"
+	tokString "DOS"
+	tokString "OLD"
+	tokString "GEOS"
+	tokString "VPOKE"
+	tokString "VLOAD"
+	tokString "SCREEN"
+	tokString "PSET"
+	tokString "LINE"
+	tokString "FRAME"
+	tokString "RECT"
+	tokString "CHAR"
+	tokString "VPEEK"
+	tokString "BANK"
 	.byt 0
 num_esc_statements = 12
-num_esc_functions = 1
+num_esc_functions = 2
 ;**************************************
 
 err01	.byt "TOO MANY FILE",$d3
--- a/basic/token2.s
+++ b/basic/token2.s
@@ -43,6 +43,7 @@ onefun	=$b4
 	.byt "TA",$ce
 	.byt "AT",$ce
 	.byt "PEE",$cb
+fnpeek	=$c2
 	.byt "LE",$ce
 	.byt "STR",$a4
 	.byt "VA",$cc
--- a/basic/tokens.s
+++ b/basic/tokens.s
@@ -1,7 +1,11 @@
+	.include "macros.inc"
+
 	.segment "BASIC"
 	jmp init         ;hard reset
 	jmp panic        ;soft reset
+
 	.byt "CBMBASIC"
+
 stmdsp	.word end-1
 	.word for-1
 	.word next-1
@@ -98,49 +102,50 @@ stmdsp2	; statements
 	.word char-1
 	; functions
 	.word vpeek
+	.word fnBank
 ;**************************************
 
-reslst	.byt "EN",$c4
+reslst	tokString "END"
 endtk	=$80
-	.byt "FO",$d2
+	tokString "FOR"
 fortk	=$81
-	.byt "NEX",$d4
-	.byt "DAT",$c1
+	tokString "NEXT"
+	tokString "DATA"
 datatk	=$83
-	.byt "INPUT",$a3
-	.byt "INPU",$d4
-	.byt "DI",$cd
-	.byt "REA",$c4
-	.byt "LE",$d4
-	.byt "GOT",$cf
+	tokString "INPUT#"
+	tokString "INPUT"
+	tokString "DIM"
+	tokString "READ"
+	tokString "LET"
+	tokString "GOTO"
 gototk	=$89
-	.byt "RU",$ce
-	.byt "I",$c6
-	.byt "RESTOR",$c5
-	.byt "GOSU",$c2
+	tokString "RUN"
+	tokString "IF"
+	tokString "RESTORE"
+	tokString "GOSUB"
 gosutk	=$8d
-	.byt "RETUR",$ce
-	.byt "RE",$cd
+	tokString "RETURN"
+	tokString "REM"
 remtk	=$8f
-	.byt "STO",$d0
-	.byt "O",$ce
-	.byt "WAI",$d4
-	.byt "LOA",$c4
-	.byt "SAV",$c5
-	.byt "VERIF",$d9
-	.byt "DE",$c6
-	.byt "POK",$c5
-	.byt "PRINT",$a3
-	.byt "PRIN",$d4
+	tokString "STOP"
+	tokString "ON"
+	tokString "WAIT"
+	tokString "LOAD"
+	tokString "SAVE"
+	tokString "VERIFY"
+	tokString "DEF"
+	tokString "POKE"
+	tokString "PRINT#"
+	tokString "PRINT"
 printk	=$99
-	.byt "CON",$d4
-	.byt "LIS",$d4
-	.byt "CL",$d2
-	.byt "CM",$c4
-	.byt "SY",$d3
-	.byt "OPE",$ce
-	.byt "CLOS",$c5
-	.byt "GE",$d4
-	.byt "NE",$d7
+	tokString "CONT"
+	tokString "LIST"
+	tokString "CLR"
+	tokString "CMD"
+	tokString "SYS"
+	tokString "OPEN"
+	tokString "CLOSE"
+	tokString "GET"
+	tokString "NEW"
 scratk	=$a2
 
--- a/basic/x16additions.s
+++ b/basic/x16additions.s
@@ -240,3 +240,20 @@ disk_done
 	lda #LOGADD
 	sec
 	jmp close
+
+;***************
+fnBank
+	lda membank
+	sta poker
+	jsr chrget
+	jsr chkopn ; open paren
+	jsr chrgot
+	cmp #')'
+	beq :+
+	
+	jsr getbyt ; byte: bank
+	stx membank
+
+:	jsr chkcls ; closing paren
+	ldy poker
+	jmp sngflt
--- a/kernal/init.s
+++ b/kernal/init.s
@@ -47,12 +47,11 @@ vectse
 
 ; ramtas - memory size check and set
 ;
-ramtas	lda #0          ;zero low memory
-	tay
-ramtz0	sta $0000,y     ;zero page
-	sta $0200,y     ;user buffers and vars
-	sta $0300,y     ;system space and user space
-	iny
+ramtas	ldx #0          ;zero low memory
+ramtz0	stz $0000,x     ;zero page
+	stz $0200,x     ;user buffers and vars
+	stz $0300,x     ;system space and user space
+	inx
 	bne ramtz0
 ;
 ; set top of memory
@@ -69,13 +68,48 @@ ramtz0	sta $0000,y     ;zero page
 ; copy banking code into RAM
 ;
 .import __KERNRAM_LOAD__, __KERNRAM_RUN__, __KERNRAM_SIZE__
-.assert __KERNRAM_SIZE__ < $0100, error, "KERNRAM size overflows one page"
-	ldx #<__KERNRAM_SIZE__
-ramtz1	lda __KERNRAM_LOAD__-1,x
-	sta __KERNRAM_RUN__-1,x
-	dex
-	bne ramtz1
+; .assert __KERNRAM_SIZE__ < $0100, error, "KERNRAM size overflows one page"
+; 	ldx #<__KERNRAM_SIZE__
+; ramtz1	lda __KERNRAM_LOAD__-1,x
+; 	sta __KERNRAM_RUN__-1,x
+; 	dex
+; 	bne ramtz1
 
+	lda #<__KERNRAM_LOAD__
+	sta $00
+	lda #>__KERNRAM_LOAD__
+	sta $01
+	lda #<__KERNRAM_RUN__
+	sta $02
+	lda #>__KERNRAM_RUN__
+	sta $03
+	lda #<(__KERNRAM_SIZE__+1)
+	sta $04
+	lda #>(__KERNRAM_SIZE__+1)
+	sta $05
+move:
+	lda ($00)
+	sta ($02)
+
+	inc $00
+	bne :+
+	inc $01
+:
+	inc $02
+	bne :+
+	inc $03
+:
+	dec $04
+	bne move
+	dec $04
+	dec $05
+	bpl move
+
+; cleanup
+	ldx #5
+:	stz $00,x
+	dex
+	bpl :-
 ;
 ; detect number of RAM banks
 ;
@@ -114,7 +148,7 @@ ioinit
 	lda #$ff
 	sta d1ddra
 	sta d1ddrb
-	lda #0
+	ina
 	sta d1pra ; RAM bank
 ;
 ;jsr clkhi ;clkhi to release serial devices  ^
--- a/kernal/routines.s
+++ b/kernal/routines.s
@@ -1,5 +1,5 @@
 .export jsrfar, banked_irq
-.export fetvec, fetch
+.export fetvec, fetch, stavec, stash
 
 	.segment "ROUTINES"
 
--- a/kernal/routines.s
+++ b/kernal/routines.s
@@ -367,7 +367,7 @@ fjmpbank
 
 	; this should not live in the vector area, but it's ok for now
 restore_basic:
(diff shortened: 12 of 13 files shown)


kktos:
**I must say it's more a hack than anything else.**

_As a side note, someone raised that SYS is calling BASIC bank when PEEK is getting values from KERNAL bank.
incoherence._

mobluse:
It is ROM bank or RAM bank the parameter is changing?

mist64:
I'd rather have a new BANK<n> command that changes the bank used for PEEK, POKE and SYS. :/

kktos:
> It is ROM bank or RAM bank the parameter is changing?
That's a pretty good question ! PEEK is using FETCH from kernal which set both bank before reading.
So to answer your question, the parm is changing both banks, read the value and set both banks to their original values

kktos:
@mist64 
Are we having one setting, one value for both banks, ROM & RAM ?
And where to store the value ?

kktos:
I've expanded by 1 byte the BVARS segment and to not move the already exising variables, I expanded it from the beginning:
`BVARS:    start = $02CC, size = $0034; # BASIC`

EgonOlsen71:
I guess BANK() returns just the current bank while BANK(n) sets a new and returns the old one? Just note that this breaks with the convention of how functions are defined in Commodore BASIC. There is no other function that changes it's semantics based on the parameter count nor is there any function that works with an empty parameter. Even FRE has to have a parameter even if it's not used for anything.

kktos:
@EgonOlsen71 Thx for the info, I didn't know. Do you have an idea to fix this ?
While doing it, I was thinking it was a bit overkill to have a function only to change a system setting. If we decided a location (in main RAM, of course) for this setting, POKE and PEEK would be enough. But less explicit, I reckon.
Could a system variable could be used ? e.g: _BANK

EgonOlsen71:
Also, there is no function that actually changes the state of something. A function is by definition supposed to have no side effects (ignoring that RND(0) of course changes the state of the random number generator). No functional programming language would work if that would be the case.

To fix this, I would think that setting the bank has to be a command, not a function with a side effect. Something like SETBANK <n> and GETBANK would have to be a function with a (useless) parameter to return the current bank. I see the idea behind BANK as it is now, but letting a function set a value in memory and return another one is bad design IMHO.

kktos:
@EgonOlsen71 okay, we're on the same page here.
But I still think it's overkill to have a specific call to set an unique setting. So to follow your proposal, what about `SET settingName, value` and `GET(settingName)`.
It seems more powerful as it can be used for other settings.
And as you are on strict mode (fine with me ;)), this would avoid a function call with a useless parm which is being far from elegant.

EgonOlsen71:
@kktos I'm not sure if I would like that. It's a bit too generic for my taste. It reminds me of what I like to call the lazy context trap: If we don't know where to put someting...just create a generic "context" and put it there. Or call it "config" if you like. I should know, I'm doing it all the time...
To be honest, I tend to like your first (PEEK bank, addr) solution better. Enhance that with a POKE bank, addr, val command and at least it's the same syntax as VPEEK and VPOKE. I don't see why the video chip's banks are addressed in one way and the cpu's banks has to be addressed in another!? The only thing that this couldn't handle is SYS into a bank, but I would argue that this is such a rare use case that it doesn't really matter. Nobody who has the skills to write something that requires this options can't write a little piece of stub assembly code to do the setup and the call that way. I don't think that BASIC needs support for this.

kktos:
@EgonOlsen71 so you're suggesting back to square one ;) fine by me :)
@mist64 , what do you think of it ? Could we be able to change your mind on this ?

For SYS, keep in mind we have many ROM banks and we may want to call some functions in a specific bank, not only in the BASIC.
And for the RAM, we may have some libs loaded here that we can use in our BASIC progs, like some graphical one, for instance.

EgonOlsen71:
@kktos If having SYS to be aware of some bank setting, I would go for a SETBANK command and a GETBANK function. However, my argument above was that people who are using SYS to call something in some ROM bank, should be advanced enough to write themselves a little piece of proxy code in normal RAM that handles these calls and for the RAM, you can always use the normal POKE to switch to the proper RAM bank. Sure, it's not "nice" but calling some routines via BASIC using SYS isn't nice anyway...

kktos:
@EgonOlsen71 hum.....
you were suggesting to go back to the initial proposal, that is `PEEK(bank, address)`.
So to be coherent, we could have a `SYS bank, address`.
but let's think a bit further here....

Let's say the band of brothers upgrade the cpu for the 65816 at some point. So we will have 24bits addresses. So why not using already 24bits addresses ?
Like a PEEK($04C000), POKE $04C000, $60 and SYS $04C000.
A bit trickier (!!) to implement but way more elegant, don't you think ?

EgonOlsen71:
@kktos IIRC, even with the 65816, they wouldn't use the extended address range. A SYS bank, addr would be an option as well. You have to keep in mind that I'm looking at this stuff from a slightly different perspective: I actually don't care much how the interpreter handles this but what it means for a compiler. Using 24bit addresses would render some peek/poke optimizations impossible and slow things down quite a lot in compiled code, so...if you don't need to have it and it actually doesn't reflect the machine's architecture, anyway even if you are using a 65816, then I would still prefer the bank, addr way of doing it. Because, after all, that's what you do for the video chip as well.

kktos:
So right now, we have :

- **BANK()**
return current bank
- **BANK( bank )**
set current bank and return previous value
- **SYS addr**
call addr; if addr>=$A000 then use bank value
- **PEEK( addr )**
get value from addr; if addr>=$A000 then use bank value
- **POKE addr, value**
set value to addr; if addr>=$A000 then use bank value

**@EgonOlsen71** doesn't like function with side effect and function w/o parm as it is not coherent with cbm basic.
Proposal is: 
- **GETBANK( dummy )**
- **SETBANK bank**

I guess we can even have this:
- **BANK( dummy )**
- **BANK bank**

And finally, we have also this possibility (_@mist64  doesn't like it_)
- **SYS [bank, ] addr**
- **PEEK( [bank, ] addr )**
- **POKE [bank, ] addr, value**

EgonOlsen71:
Making BANK a command AND a function (as in your second example), isn't a good idea either, because then you can't decide anymore if the function or the command is used without looking at the context.
I still think that the last suggestion (bank as part of SYS, POKE and PEEK) is the best solution.

Reasons:

1. It's the same syntax that VPOKE and VPEEK are using. Why use two different approaches for what's basically the same thing?
2. All these commands require at least an address. This address is defined by the bank AND the actual address. It makes no sense to me to make one part explicit in the command/function (the addr in this case) while the other one (the bank) is hidden in some internal state.
3. Using ...BANK instead of making the bank explicit in the call requires the program to manage banking state, which is actually completely unnecessary. For example, if your program uses a subroutine that POKEs into some bank, it has to save the previous bank and restore it afterwards. That's not needed at all, if the bank would a part of POKE. An example (copy some data from one bank to the other):

```
1000 REM COPY SUBROUTINE (EXPLICIT)
1010 SB=1:TB=3: REM DEFINE TARGET AND SOURCE BANK
1020 FOR I=$C000 TO $C100:POKE TB, I, PEEK(SB, I):NEXT
1030 RETURN 
```

and the same thing with SETBANK/GETBANK:

```
1000 REM COPY SUBROUTINE (IMPLICIT)
1005 OB=GETBANK(0): REM STORE THE OLD BANK
1010 SB=1:TB=3: REM DEFINE TARGET AND SOURCE BANK
1020 FOR I=$C000 TO $C100
1030 SETBANK SB
1040 V=PEEK(I)
1050 SETBANK TB
1060 POKE I,V
1070 NEXT: SETBANK OB: RETURN
```

I fail to see how the BANK-version wouble be beneficial in any way.

kktos:
@EgonOlsen71 Pretty valid point, mein Herr !

the only point I disagree with is the bank/bank().
No need to check the context to determine what is what as we have really 2 different opcodes here:
BANK
and
BANK(
I tend to prefer to avoid the get/set prefix on all functions/methods when the meaning is obvious enough.
Anyway, I won't fight over it, I'm pretty flexible from time to time ;oD

mist64:
My opinion: :D

* BANK bank
set current bank and return previous value
* SYS addr
call addr; if addr>=$A000 then use bank value
* PEEK( addr )
get value from addr; if addr>=$A000 then use bank value
* POKE addr, value
set value to addr; if addr>=$A000 then use bank value

Reasoning:
* Arguments of BASIC statements are not in parentheses, so BANK is just followed by a number.
* No BANK() function, because I don't see a need for it. The system will never change the BANK, so it would only return what was set last.

mist64:
Btw, related: https://github.com/commanderx16/x16-rom/issues/17

kktos:
@mist64 okay. And what about the points raised by @EgonOlsen71 : 
- VPOKE/VPEEK using a different syntax that POKE/PEEK (incoherence)
- the ease of having POKE/PEEK holding the bank parm when doing memcopy, for instance

EgonOlsen71:
@mist64 BANK without parentheses is a command. A command doesn't return anything. A function does, but making it a function requires parentheses...and introduces the side effects that's I've mentioned above. It's bad language design to let a function set some internal state and it's equally bad to make a command return some value in addition to it's actual purpose. I know that BASIC V2 isn't particulary well designed anyway, but at least it's consistent. We shouldn't break that for no reason IMHO.

Also, if BANK is a command but returns a value, how would that look like? Would A=BANK 1 be valid? I guess so. But since it's a command, BANK 1 (without the assignment) would be valid as well?
Looking at it from a compiler/grammar based point of view, this is just wrong because it totally breaks with the way the rest of the language works (for no reason).

Apart from all that: I still don't see the point in separating the bank and the address for these commands/function. They belong together. The next level would be:

```
10 BANK 1: ADDRESS 1024: VALUE 3: POKE
```

...

EgonOlsen71:
Oh, and another reason to make the bank explicit in the call: It's easier to understand. It's totally clear what

```
10 POKE 2, $C000, 12
```

does. Not so much, if you have something like this, where you actually have to trace the code to see which banks line 30 and line 40 are actually addressing:

```
10 BANK 2: POKE $C000,1
20 GOSUB 1000
30 POKE $C000,2
30 GOSUB 2000
40 PRINT PEEK($C000)
50 END
1000 BANK 1:POKE $C001,123:RETURN
2000 BANK 4:POKE $C002,124:RETURN
```

EgonOlsen71:
@tomxp411 We could use something like BPOKE and BPEEK then instead. That wouldn't help with performance of these particular operations, of course, but at least it won't affect POKE und PEEK in their default forms. 
However, I had too much time cycling back home, so I came up with two additional points why BANK isn't a good idea...sorry, for that...;-)

1. Using PEEK in more complex expressions. For example, if you have two banks with data, want to do a calculation on these values and store the result in another one, you could do this:

```
10 DEF FN AVG(X)=INT ((PEEK(1,X)+PEEK(2,X))/2)
20 FOR I=$A000 TO $A500: POKE3,I,FN AVG(I):NEXT
```

With BANK, that's not possible and you would have to do this:

```
10 FOR I=$A000 TO $A500:BANK 1:V1=PEEK(I)
20 BANK 2:V2=PEEK(I):V3=INT(((V1+V2)/2)
30 BANK 3:POKE I, V3:NEXT
```
Sure it works, but...come on...it's ugly as hell.

2. Again...consistency: Apart from VPOKE and VPEEK, there are also VLOAD and LOAD, which take a BANK parameter. Even your suggested COPY/VCOPY commands take two BANK parameters. It requires an extra explaination in the dcos why BANK only affects some statements and not others and that it actually doesn't set a bank, as the name might imply.

So...the only pro-BANK argument that I've heard is performance. I get that, but I don't think that it should be the one that tops everything else. I would rather use different BPOKE and BPEEK commands for accessing the banks and live with the worse performance then introducing an inconsistent and confusing BANK statement.

kktos:
I like the BPOKE/BPEEK/BSYS idea.
Better than modifying the already existing/well know commands/function.

As for the performances, hum, that's interesting.
so we could have the bank parm optional, for faster exec.
```
10 BANK 2
20 BPOKE $A000, $60 : REM using defined bank 2
30 BPOKE 1, $A000, $60 : REM using explicit bank 1
```

tomxp411:
> btw, when using a const, how basic is storing the value ? ascii or integer ? coz if it's ascii, there's a way for optimization here

BASIC doesn't have constants. You probably mean "literal." Literals are stored in ASCII characters, and the interpreter re-parses the string to a 40 bit float every time it's accessed in the code. There's a huge opportunity for optimization by simply pre-parsing all numeric literals when the program is tokenized: Michael is already aware of that, and if there's not an issue open for this, I'll add it. 

Re: BPOKE, etc. The "B" prefix in BASIC usually means "Binary"  and is used with BLOAD and BSAVE to allow for specifying a target address and range. Aside from that, I don't see any particular down side to duplicating PEEK and POKE with a "B" prefix. 

It's worth nothing that we should not add new commands with a different _suffix_, since the interpreter's string parsing is super dumb and will always take the first version of a command as it sits in the token table. So POKE vs POKEB is out of the question, since the interpreter would never parse POKEB. It will always find POKE first, since POKE is both shorter and first in the token table.  A revision to the token parsing code, to always require a space after a keyword, would fix this, but that then causes new problems. I'm happy to talk to Michael about the parsing logic in my BASIC converter as another conversation. 

The other issue is that BPEEK does nothing to reduce performance concerns, and if you cached the last used bank for a later PEEK command, that's even less clear than an explicit "set the bank now" command. The entire purpose of my proposal was to increase BASIC performance when dealing with blocks of memory, and a 3 argument peek goes in the wrong direction, slowing things down.

tomxp411:
Okay, I have some new information. I'm still testing this, but I can confirm that VPOKE is not actually 60% slower than POKE.  It's around 12% slower.

I just wrote a test program that iterates through 2048 VPOKE statements, with 8 lines of VPOKE. It then does the same thing with POKE.

`10 TI$="000000"
15 POKE $9F20,2:POKE $9F21,5:POKE $9F22,$20
20 FOR I=0 TO 255
30 POKE $9F23,I
31 POKE $9F23,I
32 POKE $9F23,I
33 POKE $9F23,I
34 POKE $9F23,I
35 POKE $9F23,I
36 POKE $9F23,I
37 POKE $9F23,I
40 NEXT
100 T1=TI
200 TI$="000000"
210 FOR I=0 TO 255
220 VPOKE 0,$0000,I
221 VPOKE 0,$0002,I
222 VPOKE 0,$0004,I
223 VPOKE 0,$0006,I
224 VPOKE 0,$0008,I
225 VPOKE 0,$000A,I
226 VPOKE 0,$000C,I
227 VPOKE 0,$000E,I
240 NEXT
250 T2=TI
300 PRINT "POKE: ";T1/60
310 PRINT "VPOKE: ";T2/60
320 PRINT "DIFF:  ";T2/T1
1000 POKE $9F20,2`

Yes, the 8 POKE/VPOKE statements in a row are redundant. This is to make sure the FOR loop is not the bottleneck.

In my original version of the program, I just VPOKEd 0,0,I, and the VPOKE was actually *faster*. Using a 4-digit number definitely increases the runtime of the statement.  To test that the integer parsing is the culprit, I changed the VPOKE lines like so:

VPOKE $0000,$0000,I

and the performance went down to about 40% slower than the POKE loop. 

Finally, I changed the code to use a variable:
POKE P,I
VPOKE B,P,I
with appropriate initializations.

Anyway, the short version is... VPOKE comes out at 12-30% slower, based on the input values.  In doing further tests with integer parsing, it looks like parsing parameters is by far the biggest time sink in BASIC commands, and so reducing the need to parse redundant parameters is probably the best approach.

tomxp411:
Note that I am not against BPEEK/BPOKE, if those are done in addition to the bank command. I just do not want to see a 3-argument poke and 3-argument peek as the only way to address banked memory, due to the performance hit.

So the following two lines should functionally do the same thing, but the first will be faster if you have to iterate through hundreds of POKE or PEEK statements:

BANK 1:POKE $A000,123
BPOKE 1,$A000,123

And sorry to hammer you with replies, Michael. I just feel like handling banks in BASIC is a super critical thing, and it needs special care. :)

kktos:
Tom, that's very nice we're able to have exchanges about that. Some of the best ideas came from synergies ;)

EgonOlsen71:
...but you can already handle, at least RAM banks, in BASIC without any performance hit by simply poking the bank into the corresponding VIA address and then POKE and PEEK to your liking. And if this is only about performance, then why not bench this way:

```
10 TI$="000000"
15 POKE $9F20,2:POKE $9F21,5:POKE $9F22,$20
20 FOR I=0 TO 255
30 POKE 40739,I
31 POKE 40739,I
32 POKE 40739,I
33 POKE 40739,I
34 POKE 40739,I
35 POKE 40739,I
36 POKE 40739,I
37 POKE 40739,I
40 NEXT
100 T1=TI
200 TI$="000000"
210 FOR I=0 TO 255
220 VPOKE 0,0,I
221 VPOKE 0,2,I
222 VPOKE 0,4,I
223 VPOKE 0,6,I
224 VPOKE 0,8,I
225 VPOKE 0,10,I
226 VPOKE 0,12,I
227 VPOKE 0,14,I
240 NEXT
250 T2=TI
300 PRINT "POKE: ";T1/60
310 PRINT "VPOKE: ";T2/60
320 PRINT "DIFF: ";T2/T1
1000 POKE $9F20,2
```

...and all of a sudden, VPOKE is much faster (because of the simpler parsing). BTW: The compiled version of this test shows no measurable difference between the two approaches, but adding BANK awareness to POKE/PEEK would slow it down in general, so from my performance point of view, I would have to be against it and pro-bank-parameters instead as well

But anyway, I think that I've said all that I could. To me, BANK is just an ugly hack for all the reasons mentioned above. Its one and only benefit is "performance" so to me, it's something like 4 to 1 against it. If it's going to become a reality, then so be it...but everything in me screams "wrong" here.

The final question is: Who is going to decide this? Somebody has to..

tomxp411:
@EgonOlsen71 this your biased test doesn't reflect the usage in banked RAM, and talking about compiled code is a non-sequiter. 

You can't just POKE the VIA, because the memory map says that bank $FF is used by the BASIC interpreter. So if you change the address in the VIA, this will crash the interpreter (assuming this information is correct.)

What matters to me is performance of the interpreter and my interpreted BASIC programs. And the simple fact is that in every real use case I've come up with, including some real world testing, it will be faster to have a dedicated BANK command and have SYS, PEEK, and POKE follow the bank number. 

If that's not a priority to you, then 

The actual ASM code to test the bank is just two compare/branch statements (with an LDA/STA pair to actually change the value if needed), and this uses far less CPU time than parsing additional parameters. 

And regardless, PEEK and POKE with the bank parameter would have to do the ROM/RAM check (to set the appropriate VIA registers), so adding parameters to PEEK and POKE is only a net loss in performance. 

I intend to write RPGs and adventure games that will use up large amounts of banked memory. I simply cannot do those in BASIC with the current state of the system's performance. This is why I requested block operations, along with string, 16-bit and floating point PEEK and POKE. The BANK keyword is part of a much larger picture that gets more complicated and much slower if it gets mucked up by adding parameters and duplicate keywords for 8 different block memory commands.

greg-king5:
`PEEK(), POKE, SYS, USR(), and WAIT`

Don't forget USR() and WAIT.

tomxp411:
I forgot about USR(). I'd probably argue that USR is out of scope for this, since that just uses a vector in BASIC RAM ($0311–$0312). That can be extended to a 3-byte address (and should be.) 

WAIT is definitely in scope and should honor the selected bank number.

EgonOlsen71:
@tomxp411 Don't know...I poked into the banks by poking into the VIA and had no issues with it in a simple test case. But that might be because of my simple test case. I didn't know that the interpreter is supposed to use one of the ram banks.
Me talking about compiled code is as valid as you talking about your RPG games using the interpreter... a change that's good for what you have in mind might be bad for what I have in mind. I'm looking at these changes, as mentioned earlier, more from a language point of view. I don't really care if something is 10 or 20% slower or faster in the interpreter but I do care about the language's consistency. If that's the opposite for your, then that's fine to me. But in the end, somebody has to decide which way to go...I'm missing a sense of direction when it comes to language changes like this. Who wins? Michael? The one implements it first? The one who favours performance? The one who favours the least intrusive solution? I just don't know, but I would like to because I have to reimplement all these changes as well.

kktos:
Egon, Michael is the owner of this project so I may guess he is the one taking the final decision. So the best way to "push" him in the right direction is doing precisely what you've done so far.
You raised very good points with neat explanations. Tom is doing the same. And at the end, I think you're both aiming in the same direction as I guess Tom would enjoy having a good basic compiler for his games ;)
I'm really delighted to see such interest and passion here. I would love to see similar behaviours at work around our code ;) So, guys, I thank you very much as you're making me smile !
Vielen Dank.

EgonOlsen71:
> Egon, Michael is the owner of this project so I may guess he is the one taking the final decision. 

Maybe...but then again, it's David's project with a mission statement to make programming on it an experience close to the C64. So he should have an opinion about what's a fitting change and what isn't. They have to produce a physical manual for it after all, so at some point there has to be a feature freeze...this discussion doesn't really belong here, it's just something that I'm wondering about.

mobluse:
Maybe one could have `BANK arg` that would change both RAM and ROM bank to `arg`, and `BANK argram,argrom` that would change banks to different values, and `BANK ,argrom` that would only change ROM bank, and `BANK argram,` that would change only RAM bank. BANK would change banks when using POKE, PEEK(), SYS, USR(), and WAIT.

I thought first empty arguments looked strange, but RENUMBER works this way so the BASIC V7 parser must be able to handle this.

One could also change VPOKE and VPEEK so that they don't use bank argument, but that the BANK command has a third argument for video bank.

tomxp411:
> One could also change VPOKE and VPEEK so that they don't use bank argument, but that the BANK command has a third argument for video bank.

If you're going to do more than 2 parameters for bank, then the way to handle that would be something like  

BANK _bankNo_ [ , _bankType_ ]
Where _bankType_ is 0, 1, or 2, for RAM, VERA, and ROM and _bankNo_ is the number to set. 

So typing BANK 1 would set the RAM bank to 1. Typing BANK 0,1 would set VERA's bank to 0. 

That would actually make VPOKE and VPEEK faster, since they would not need to parse the extra parameter. It would also unify the syntax of all of the memory access commands.

greg-king5:
`VPOKE` and `VPEEK()` _don't_ have a bank parameter!  They have an address high-byte.  That byte is separate -- at this time -- only because of a limitation in the current BASIC ROM.  If 24-bit address parsing is implemented in the interpreter, then `VPOKE` and `VPEEK()` can look like the current `POKE` and `PEEK()` -- but, `BANK` has no connection to them.

tomxp411:
The syntax for the VPOKE command is 

**VPOKE**
TYPE: Command FORMAT: VPOKE <bank>, <address>, <value>

This is taken straight from the Programmer's Reference Manual. So you can take up the nomenclature with the person who wrote the manual. But until it changes, we will keep calling it "bank", since that's what the manual calls it.

greg-king5:
> I first thought that empty arguments looked strange, but RENUMBER works this way, so the BASIC V7 parser must be able to handle this.

Microsoft's Micro BASIC always has worked that way!  `PRINT`, `NEXT`, `LIST`, and `RUN` have optional arguments.  `RESTORE`, `END`, `STOP`, and `CONT` have no arguments.

BruceMcF:
> ...but you can already handle, at least RAM banks, in BASIC without any performance hit by simply poking the bank into the corresponding VIA address and then POKE and PEEK to your liking.

That's what BANK could well be doing ... just syntax sugar for the magic number POKE.

Basic "using" segment $FF would then be a routine that uses that side RAM stacking the current BANK, changing to $FF, then restoring the current BANK when it's done. So SYS "calls the current bank" without any change to the original SYS, PEEK "uses the current bank" without any change to the original PEEK, POKE "uses the current bank" without any change to the original POKE.

There is an argument to flipped and is being EMULATED while leaving $FF as the "normal" value, because if you wrap PEEK, POKE and SYS with a copy of a stored "pseudo BANK register", you don't need to stack the prior BANK register value ... it's $FF, just store that when you are done. Especially if $FF is used by the main keyword interpreter loop and not just by specific routines, this will be more efficient, because every keyword would need the $FF segment and only three would need to access the psuedo bank.

But if implementing a psuedo-BANK register, you ought to go all the way. The REAL BANK register in the VIA can be read: it is NOT a blind latch. So if using a pseudo-BANK-register, it should be able to be read.

It reduces errors in code to retrieve the actual state rather than tracking what the state "is supposed to be". In other words, storing the initial value in a variable because you are writing to a blind latch is one thing, but designing something to be a blind latch when there is no need to do so is poor language design. The benefit of using a direct report of system state only increases when programming in teams or when one person is working on code someone else wrote.

Now, it's not critical to have a keyword for that ... in the design with "the BANK is the actual BANK, Basic accesses $FF behind the scenes" that this approach is emulating, BANK is just a stable location accessible with PEEK and POKE. So a stable location in $03xx for the psuedo-BANK value would be fine, and then "BANK number" is syntax sugar for storing a value into it.

High RAM and ROM Banks are different. The original basic PURPOSE of High RAM was to give Basic additional room to store data. Those PEEK and POKE ought to be done efficiently, as they might be done 8K times in a tight inner FOR/NEXT loop.

But Basic is running from ROM. SYS and PEEK and POKE ought to "see" the same address, and having the SYS call a different ROM is a substantial increase in complexity over what is required for call to a routine in Low RAM or High RAM segment.

SO:
  FARSYS address, rombank
  data=FARPEEK(address, ROMBANK)
... default ROMBANK is the Kernel bank ... would be appropriate for alternative ROMBANK access.

Alternatively, IF it happens that it doesn't increasing the parsing overhead for a function keyword when there is an optional second parameter but it's skipped, then:

data=PEEK(address [,rombank])
FARSYS address [, rombank]

kktos:
so far, here is what I gathered from our conversation:

- **`BANK number : REM 1st proposal`**
set the default bank number (RAM&ROM) used by BPOKE/BPEEK/BSYS/BWAIT/BUSR
- **`BANK ramnumber [ ,romnumber ] : REM 2nd proposal`**
set the default bank number for RAM and optionally ROM  used by BPOKE/BPEEK/BSYS/BWAIT/BUSR
- **`BPEEK([bank,] address)`**
read value at address using either default bank number (may be diff if either ram or rom) or the given bank number
- **`BPOKE [rambank,] address, value`**
store value at address using either default bank number (ram only, obviously) or the given bank number
- **`BSYS [bank,] address`**
call address using either default bank number (may be diff if either ram or rom) or the given bank number
- **`BWAIT [rambank,] address, mask1 [,mask2]`**
Wait for address using either default bank number (ram only, obviously) or the given bank number
- **`BUSR([bank,] value) : REM 1st proposal`**
call the userdef fn address using either default bank number (may be diff if either ram or rom) or the given bank number
- **`BUSR(value) : REM 2nd proposal`**
Same here but we have to reserve another memory address to store the bank (
userdef fn address: $311-$312; bank: ???)

I leave BCOPY/BMOVE out of this list for now but that will be the next step.

mobluse:
I don't agree that the address commands should be preceeded by B and have a bank parameter. E.g. in GW-BASIC (a later Microsoft BASIC after Commodore) they didn't need this: http://www.antonis.de/qbebooks/gwbasman/defseg.html
One could have DEF ROM and DEF RAM for ROM and RAM bank, respectively, otherwise one of the BANK syntaxes will do. It might be sensible to POKE into ROM since that could be used for output to accessories.

kktos:
Mikael, I’m trying to gather all suggestions here and one of them was that the bank should be there with pretty good reasons (cf Egon’s posts)
Another was saying to alway pass the bank would have a performance hit.
Hence the choice to pass it or not.

Kroc:
A suggestion, but nothing I'm attached to:

* `BRAM` set RAM bank
* `BRAM()` get RAM bank
* `BROM()` get ROM bank

kktos:
@mist64 What the status, Kenneth ? ;) What do you want me to do ? I'm all ears.

BruceMcF:
Assume Version 1: BANK N. It stores the "HighRAM bank value", HR_VAL. Document where "HR_VAL" is located in Low RAM, don't move it around, and someone who wants to read the current target BANK can peek it.

Setting the Ram BANK doesn't need any address parsing, and won't need springboarding of the call in low RAM, just wrap each of the PEEK, POKE, SYS USR and WAIT with "LDA HR_VAL; STA HR_REG; ...; LDA HR_BASIC; STA HR_REG", done. So BANK N just needs to store a value in "HR_VAL".

PEEK, POKE, SYS, USR and WAIT now work with the current emulated High RAM bank setting, and take the ROM Bank setting as the Basic ROM.

ROM versions of WAIT and POKE are not needed, so they are done.

There is no particular reason to have a "Cross ROM" USR call. If the USR function wants to play games with the ROM bank, it can do that in software ... it can, after all, be located in one of 63 to 255 High RAM segments, that is plenty of space for the USR code to include ROM springboarding if it is making use of a user defined ROM segment. So USR is also done.

ROMPEEKaddr,rom) needs springboarding. It will be executing in the Basic ROM and is getting a data byte from another ROM. Somewhere ... in a Basic Page or in the Basic High RAM segment ... is a routine in RAM which switches the ROM bank, gets the value, switches the ROM value back, and hands the byte back.

ROMSYS addr [,rom [,A,X,Y,flags]] also needs springboarding. Somewhere is a routine in RAM that switches the ROM bank, makes the call, gets the return, switches the ROM value back, and returns to Basic.

Note that this is also the version of the SYS call with the V7 style optional A,X,Y,flags parameters, so Basic helper routines in some High RAM segment can use whatever Basic parameters are sitting after the V2 "SYS" call.

kktos:
- **`BANK number`**
set the default bank number (RAM&ROM) used by BPOKE/BPEEK/BSYS
- **`BPEEK([bank,] address)`**
read value at address using either default bank number or the given bank number
- **`BPOKE [bank,] address, value`**
store value at RAM-only address using either default bank number or the given bank number
- **`BSYS [bank,] address [,A value] [,X value] [,Y value] [,S value]`**
call address using either default bank number or the given bank number with optional values for CPU registers

kktos:
> USR and WAIT

oops, you're right, I removed them from the list a tad too early ;)

> the use of HR_BASIC is hidden inside the...

This is done already in my code ;) But since there were a lot of new ideas, it's on hold for now until something is decided. I've tweaked POKE/PEEK, for example, but now, after this long conversation, I believe it wasn't the best choice. Best to leave them as is and add new commands like BPOKE & BPEEK.

> In CBM V2, how you tell that an optional value is omitted is you get to the end of the expression. So that's BPEEK (address [, ROMbank]) and BANK is only needed for ROM.

okay, you can. But it's not intuitive and I don't like it as a developer. If I had to use it, I would have wonder why the guy who wrote it did put the parms in the wrong order. Probably a lazy dev. So if I'm the one writing it, I have the same demands. Doing it right so if I have to use it, I will be pleased.
Let's be demanding :)

> Again, mandatory value comes first, options after.

That's the easy way, not the right way. So no go. :)

> BANK

I was wondering about a system variable, like TI (TIME)...... instead of a function and/or a procedure....

BruceMcF:
> > the use of HR_BASIC is hidden inside the...
 
> This is done already in my code ;) But since there were a lot of new ideas, it's on hold for now until something is decided. I've tweaked POKE/PEEK, for example, but now, after this long conversation, I believe it wasn't the best choice. Best to leave them as is and add new commands like BPOKE & BPEEK.

Better to make it consistently "as if" the VIA has been set then to have one set going to a bank depending on what the interpreter is looking at and another set depending on where the user has told them to look.

> > In CBM V2, how you tell that an optional value is omitted is you get to the end of the expression. So that's BPEEK (address [, ROMbank]) and BANK is only needed for ROM.
 
> okay, you can. But it's not intuitive and I don't like it as a developer.

In what way is it not intuitive? It's intuitive to people who use this Basic, because that's how other ones work. It's intuitive to users of SYS and PEEK and POKE because it follows their pattern and then adds stuff. It's intuitive to programmers of 65xx family processors, because that's information you'd put in base+2.
Intuition is mostly "I'm used to", and it seems like it'd be just as intuitive either way.

> If I had to use it, I would have wonder why the guy who wrote it did put the parms in the wrong order. Probably a lazy dev. So if I'm the one writing it, I have the same demands. Doing it right so if I have to use it, I will be pleased.
> Let's be demanding :)

If I thought it was the right order, that would be different. My first instinct for arbitrary decisions like this is the order that is easier for beginners to use, which is address, bank.

> > BANK
 
> I was wondering about a system variable, like TI (TIME)...... instead of a function and/or a procedure....

If variable names are extended, I'd be all for it. I had a suggestion of a system variable that I deleted, because while I like BANK% working like TIME$, I'm not sure about BA% though.

For the subroutine making use of a specific bank and returning the bank to the bank when called, the oldbank=BANK(newbank) procedure is ideal. For a main loop, it can just remember define variables for the various banks it uses and call "BANK(MAP1)", letting the "oldbank" value silently go away when it's not assigned to anything.

... hmmm, at least, AFAIR. I'll test that in the C64 V2 when I get back from morning classes.

tomxp411:
re: system variables - it's bad enough we have ST, TI, and TO already reserved... let's not add any more. Let's also not rush to add a ton of new tokens that largely duplicate existing functionality.

The BANK command is already part of Commodore BASIC. Why re-invent the wheel when we already have a template? 
image

BruceMcF:
Two significant character variables really make my skeptical about having good names. If the variable is "HR" for "High RAM", some program wants to use "HR" for the number of home runs and some program wants "HR" as Human Resource department.

The BANK token is there, but the BANK "template" from the 128 would have us defined a single number that specifies both ROM and RAM bank, EG, bank number is 256*rombank+rambank. But the proposals have never had a single bankvalue that dictates both at the same time, the way that C128 bankvalues work.

So "the template is already there" is only being used in a loose sense, and exactly how to interpret it in the context of the CX16 if using BANK is still subject to what is most appropriate. We are reinventing the wheel, and changing the template ... the question is how to change it.

And what I said above at the outset ... assume proposal 1, BANK bankvalue ... follows the template very closely. What you set is what you can see when using memory access routines, so BANK is selected from {$00...$FF} and picks the RAM bank context.

 I do very much like "oldvalue=BANK(newvalue)" for the benefit of using it in a subroutine, but as I said, as long as the place where the emulated bank value is stored is at a documented, stable location so it is accessible, that would at least make it possible to write proper subroutines when a subroutine is doing it's work in a different segment than the caller.

kktos:
> It's intuitive to people who use this Basic, because that's how other ones work. It's intuitive to users of SYS and PEEK and POKE because it follows their pattern and then adds stuff. It's intuitive to programmers of 65xx family processors, because that's information you'd put in base+2.
Intuition is mostly "I'm used to", and it seems like it'd be just as intuitive either way.

That's a pretty sensible argument, if not the most. okay, then. :)

> re: system variables - it's bad enough we have ST, TI, and TO already reserved... let's not add any more. Likewise, let's also not rush to add a ton of new tokens that largely duplicate existing functionality.

yep, was just a thought after I noticed TIME & TIME$. I thought it could solve the BANK as a command, as function.....
But if those system variables were already a pain by then (C64 good-old-time), let's forget about and stick to the original plan.

>oldvalue= BANK(newvalue)

This was already discussed and put aside as "impure" function, that is with side effect. A proper function would have been like `currentValue=BANK()` but I learn we cannot have no parameter for a function (!!) so again, for consistency, this was put aside.

BruceMcF:
> > It's intuitive to people who use this Basic, because that's how other ones work. It's intuitive to users of SYS and PEEK and POKE because it follows their pattern and then adds stuff. It's intuitive to programmers of 65xx family processors, because that's information you'd put in base+2.
> > Intuition is mostly "I'm used to", and it seems like it'd be just as intuitive either way.
 
> That's a pretty sensible argument, if not the most. okay, then. :)

On the other hand, it's counterintuitive to VPOKE and VPEEK ... but if you follow VPOKE and VPEEK, the bank would not be an optional parameter.

Note that if (ram) BANK is simply wrapped around each PEEK, POKE, SYS, USR and WAIT (if the "High page" is the in context in Basic interpreter), or if (ram) BANK is simply stored in the VIA register (if the "High page" is put into context for specific routines that use it) ... mandatory BANK for BPOKE, BPEEK and BSYS works fine, and the objection to slowing down parsing of PEEK and POKE goes away because parsing of BANK is only done when it's needed. 


> > oldvalue= BANK(newvalue)
 
> This was already discussed and put aside as "impure" function, that is with side effect. A proper function would have been like `currentValue=BANK()` but I learn we cannot have no parameter for a function (!!) so again, for consistency, this was put aside.

That's more of an argument over the semantics of "function". It is indeed not a function, it's a procedure, in the style sometimes used in procedures supporting mutex operations. Procedures that return values are often colloquially referred to as functions in Basic, because they use the same syntax as functions and the defining word for them in many Basics is called "FUNCTION" (cf. [QBasic](https://en.wikibooks.org/wiki/QBasic/Subroutines_and_Functions) where "A function is a form of subroutine that returns a value" ... which is to say, it is not limited to actual functions).

It is the procedure "oldvalue = BANK(newvalue)" which is more useful than "BANK" ... but only marginally, so long as the location of the pseudo-register value set by BANK is officially documented and stable. So the latter would suffice, given that the CBM programmers decided to make BANK a statement in CBM Basic V7.

BruceMcF:
So, taking all of that into account, using the approach to BANK and the existing functions that mist64 says he prefers:

BANK rambank ... this sets the default bank used by all memory access functions. The location that the rambank value is written to is published and stable.

value=PEEK(address) ... using default bank if $A000<=address<=$BFFF

POKE address, value ... using default bank if $A000<=address<=$BFFF

SYS address ... using default rambank irrespective of address (that is, a routine at $9EF8 has the right to expect that the current default rambank is the rambank it sees when it starts).

value=USR(value) ... the USR routine uses default rambank irrespective of address

WAIT address, mask1 [, mask2] ... using default bank if $A000<=address<=$BFFF

This really covers a lot of use cases. For explicit banking access, the required functions are calling a routine in an alternative rombank, reading a value from an alternative rombank, and either peeking or poking to an alternative rambank (for copy between rambanks ... one of the two sides can use the default bank, and for greater SHOULD, use the default bank).

Using VPOKE, VPEEK analogy, so bank is not optional:

value=ROMPEEK(rombank, address) ... using rombank if $C000<=address

ROMSYS rombank, address [,A value] [,X value] [,Y value] [,S value] ...  ... using rombank if $C000<=address

BPOKE rambank, address, value, using the specified bank if $A000<=address<=$BFFF

eg,
100 SB=PEEK(HRAM): BANK M8: FOR I=$A000 TO $BFFF: BPOKE M9,I,PEEK(I): NEXT: BANK SB

BruceMcF:
> 
> 
> I don't agree that the address commands should be preceeded by B and have a bank parameter. E.g. in GW-BASIC (a later Microsoft BASIC after Commodore) they didn't need this: http://www.antonis.de/qbebooks/gwbasman/defseg.html
> One could have DEF ROM and DEF RAM for ROM and RAM bank, respectively, otherwise one of the BANK syntaxes will do. It might be sensible to POKE into ROM since that could be used for output to accessories.

But if you were poking into ROM to communicate with an accessory on an expansion card (eg, STA $FF00,X stores the word made up of the data byte and the page address to register $F)... why would you need banking? The expansion card would only know about the address, it wouldn't know about what the three address lines from the VIA are telling to the ROM.

mobluse:
The address lines from VIA are perhaps on the expansion bus, but if they are not I suppose DEF ROM could be left out.

BruceMcF:
> 
> 
> The address lines from VIA are perhaps on the expansion bus, but if they are not I suppose DEF ROM could be left out.

If the Basic extensions are looking to be parsimonious in the new tokens that they use, then one cross-rom SYS command and one copy-from-rom command suffice for commands where it is necessary to specify a bank. Since the logic of executing either a sys in a different ROM bank or a copy from another ROM bank is so different from the equivalent of any other command, it makes sense to have them as distinct keyword. And with two keywords where setting the target ROM is mandatory, it saves a keyword to include that in the parameters rather than set them separately.

So only "DEF RAM" is needed, which can be "BANK value" as v7 Basic already has a token allocated for BANK.

mobluse:
But you still need to PEEK from other ROM banks, so something like DEF ROM might be needed. I would e.g. like to make a program that prints the current keyboard layout to the screen and then I need to read other ROM banks.

BruceMcF:
> But you still need to PEEK in other ROM banks, so something like DEF ROM might be needed. I would e.g. like to make a program that prints the current keyboard layout to the screen and then I need to read other ROM banks.

Why wouldn't the mooted FARCOPY handle that ... "or a copy from another ROM bank" ... and do it better, since why copy the keyboard layout byte by byte? Something like:

"FARCOPY rombank, address1, address 2 [, value] ... address 1 is the source address, in rombank if $C000 <= address 1, value is number of bytes to copy, default 1 byte."

kktos:
@mist64 It would be nice if you can decide what to do with this. As the master branch is now quite ahead, I'll have to resolve new conflicts. you know how pleasant that is ;). So would you be so kind to tell me what to do with this, either keep it as is or drop it completely or change it for something else.
Vielen Dank.

BruceMcF:
If it has been established in the most recent revision that when dedicated Banks are used for actions of Basic, the currently visible bank is to be saved when calling that action and restored on return ... which seems to be the case when the change of Bank after Basic used the Kernel was fixed as a bug ...

... it is probably time to set BANK as storing to the RAMBANK Via register, document the effects on existing keywords, and narrow the bounds for what other capabilities ought to be handled.

IOW:
____________
BANK rambank ... this sets the default bank used by all memory access functions. The location that the rambank value is written to is published and stable. By default, "rambank"=1 {+}
{+Whether or not to have a Bank allocation system that will change the default Basic RAMBank can be TBD later.} 

value=PEEK(address) ... using default bank if $A000<=address<=$BFFF

POKE address, value ... using default bank if $A000<=address<=$BFFF

SYS address ... using default rambank irrespective of address (that is, a routine at $0700 has the right to expect that the current default rambank is the rambank it sees when it starts). By default, "address" is the address following the end of the Basic program{+}.
{+ Note that this is a facility for ML programs or ML initialization routines for PRG files that start with a Basic stub so that they can be auto-run from the Basic command line.}

value=USR(value) ... the USR routine uses default rambank irrespective of routine address stored for USR.

WAIT address, mask1 [, mask2] ... using default bank if $A000<=address<=$BFFF
_______
... and then sort out the balance at a later date.

PR #68, opened 2019-10-23
Float optimization

MJoergen:
I've rewritten the floating point addition and multiplication routines from scratch, and placed them in two separate files: basic/fadd.s and basic/fmult.s. These files contain commented assembly code.

I've removed most of the old add/mult routines from code17.s and code19.s. Some parts of the old code is still used in other math functions like INT(), SGN(), etc., and is left unchanged.

I've added a directory test, which I've used to test the correctness of the new routines, as well as do some performance benchmarking. On average, the cost of an addition has gone from 400 clock cycles to 300 clock cycles, and the cost of a multiplication has gone from 2000 clock cycles to 1500 clock cycles.


mist64:
I'm a big fan of this. :) I hope I'll have some time to have a closer look at this soon.

MJoergen:
This pull request has gone stale. Should I make a new pull request based on the latest version of the ROM ?

MJoergen:
This PR can be closed. I've updated the files and made a new PR #143 .

greg-king5:
You can close your own issues and pull requests.

Issue #79, opened 2019-11-09
The code in editor.2.s only works because of a bug in the emulator and won't work on real hardware


differentprogramming:
The "bit" instruction hack is used to skip a 3 byte instruction instead of a 2 byte instruction.
The reason it's working is that the unmasked byte is $02 and the emulator mistakenly assumes that all unused instruction are 1 byte NOPS.  In reality it's a 2 byte NOP on the 65c02.  

Not sure about a 65816 in emulation mode. In regular mode it's a 2 byte COP instruction. And on the original 6502 it's a KIL.

The original Kernal was 8k and you have 16k, so PLEASE get rid of these tricks that stop the assembler from flagging errors. You just don't need the extra bytes.  And if you really want to optimize the code, then do it after the code is stable and only make changes that will be significant.   

****************************************
; RECEIVE SCANCODE AFTER shflag
; * key down only
; * modifiers have been interpreted
; and filtered
; out: X: prefix (E0, E1; 0 = none)
; A: scancode low (0 = none)
; Z: scancode available
; 0: yes
; 1: no
;****************************************
receive_down_scancode_no_modifiers:
jsr receive_scancode
beq no_key
jsr scancode_to_joystick
php
jsr check_mod
bcc no_mod
plp
bcc key_down
eor #$ff
and shflag
.byte $2c
key_down:
;THIS LINE IS WRONG BECAUSE shflag IS NOT ON ZERO PAGE
ora shflag
sta shflag
key_up: lda #0 ; no key to return
rts
no_mod: plp
bcs key_up
no_key: rts ; original Z is retained

greg-king5:
I bet that no one recognized that problem when @mist64 moved many variables from the zero page!  Some `bit`-skips are invalid, now.  That might be one reason why Release 34 seems to be a regression.

BruceMcF:
65816 "emulation mode" is not emulating a 65C02 processor, it is emulating 65C02 opcodes ... so any opcodes that are not implemented in the base 65C02 perform their 65816 operation in emulation mode. If there is a "COP" encountered, it would execute a vectored coprocessor call whether it is in emulation mode or not.

After all, that's how you escape emulation mode ... by using a 65816 opcode that is not a base 65C02 opcode.

differentprogramming:
Found another bit instruction bug.
I'm gonna get rid of all of them and do a pull request

Look at this, a bit instruction skipping to a bne.
That can NEVER work

;  **********************************************
[...]

PR #81, opened 2019-11-11
got rid of $2c partial instruction skip

differentprogramming:
To fix various problems I listed under issue #79 I created this pull request.

The .byte $2c hack is a way of skipping a 2 byte instruction with while saving 1 byte over a bra, but it has 4 problems:
1) If the instruction to be skipped over is not 2 bytes long, say a variable moved off zeropage (as just happened) the final byte of the skipped instruction is revealed as a new instruction, which may do different things on different processors or on the emulator (which currently does not implement unused instructions properly).
2) flags are affected.  There is one case where there was a skip to a bne, but the bit instruction overwrites the zero flag.
3) randomish memory is read.  If an instruction aliased to a vera register, for instance, and the read changed the state, that would be a very hard bug to find
4) if the skip is not valid, the assembler won't flag an error.

Finally, the only point of this is to save space, but our code was taken from a machine with 8k roms and put on one with 16k roms, there is no need to save a few bytes per program this way.

Also, this is premature optimization.  You shouldn't put in optimizations that can break before code is finalized. 

****************************************************************
I don't have the ability to test all the code I changed, so I'm asking that this request be tested by someone capable of doing the testing.   I can verify that it fixed two bugs that I'm aware of.

Thank you.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,9 @@ endif
 AS           = ca65
 LD           = ld65
 
-ARGS_KERNAL=-g
-ARGS_BASIC=-g
-#ARGS_MONITOR=-g
+ARGS_KERNAL=-g --cpu 65SC02 
+ARGS_BASIC=-g --cpu 65SC02 
+ARGS_MONITOR=-g --cpu 65SC02 
 #ARGS_DOS=-g
 #ARGS_GEOS=-g
 
--- a/cbdos/main.asm
+++ b/cbdos/main.asm
@@ -417,11 +417,11 @@ cbdos_acptr:
 
 ; EOF
 	lda #$40
-	.byte $2c
+	bra @js60
 @acptr_nofd:
 ; no fd
 	lda #$02 ; timeout/file not found
-	sta $90
+@js60:	sta $90
 	lda #0
 	ldy save_y
 	ldx save_x
@@ -606,16 +606,16 @@ finished_with_buffer:
 ;****************************************
 set_status_ok:
 	lda #$00
-	.byte $2c
+	bra js61
 set_status_writeprot:
 	lda #$26
-	.byte $2c
+	bra js61
 set_status_synerr:
 	lda #$31
-	.byte $2c
+	bra js61
 set_status_73:
 	lda #$73
-	pha
+js61:	pha
 	pha
 	lsr
 	lsr
@@ -931,16 +931,16 @@ read_dir:
 	sbc #>10
 	bcs gt_10
 	ldx #3
-	.byte $2c
+	bra js62
 gt_10:
 	ldx #2
-	.byte $2c
+	bra js62
 gt_100:
 	ldx #1
-	.byte $2c
+	bra js62
 gt_1000:
 	ldx #0
-	lda #' '
+js62:	lda #' '
 :	jsr storedir
 	dex
 	bne :-
--- a/cbdos/main.asm
+++ b/cbdos/main.asm
@@ -21,9 +21,13 @@
 .importzp filenameptr, krn_ptr1, krn_ptr3, dirptr, read_blkptr, buffer, bank_save
 .include "common.inc"
 IMPORTED_FROM_MAIN=1
+
+.feature labels_without_colons
+
 .include "fat32.inc"
 .include "fcntl.inc"
 .include "65c02.inc"
+
 
 NUM_BUFS = 4
 TOTAL_NUM_BUFS = NUM_BUFS + 3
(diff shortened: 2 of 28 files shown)


greg-king5:
> I don't have the ability to test all the code I changed, so I'm asking that this request be tested by someone capable of doing the testing. I can verify that it fixed two bugs that I'm aware of.

Please describe those two bugs that will be repaired by this PR.

cbdos/main.asm
```diff
@@ -417,11 +417,11 @@ cbdos_acptr:
 
 ; EOF
 	lda #$40
-	.byte $2c
+	bra @js60
 @acptr_nofd:
 ; no fd
 	lda #$02 ; timeout/file not found
-	sta $90
+@js60:	sta $90
```
greg-king5: @mist64, that patch reveals another bug -- not related to this PR.
The `status` byte isn't there anymore!
There still are some places in the source code that have "hardwired" -- and most likely, now invalid -- addresses.  Another place that I noticed is in Kernal's `MOUSE` subroutine.  It steals zero-page location `$00` from us.

Makefile
```diff
@@ -8,9 +8,9 @@ endif
 AS           = ca65
 LD           = ld65
 
-ARGS_KERNAL=-g
-ARGS_BASIC=-g
-#ARGS_MONITOR=-g
+ARGS_KERNAL=-g --cpu 65SC02 
+ARGS_BASIC=-g --cpu 65SC02 
+ARGS_MONITOR=-g --cpu 65SC02
```
greg-king5: Won't the Commander X16 use a 65C02 microcontroller?

differentprogramming: It may use a 65816 which doesn't have the new bit instructions just like 65SC02 also doesn't have the new bit instructions.

cbdos/main.asm
differentprogramming: The two bugs are:
https://github.com/commanderx16/x16-emulator/issues/209
and
https://github.com/commanderx16/x16-rom/issues/79

The first bug is "On PC right alt and both ctrl keys only allow keys through every other stroke"

The second one is that in some cases where variables moved off zero page, the code is only working because this emulator incorrectly handles all instructions it has never heard of as 1 byte NOPs.

But in one case the revealed bug instruction is a two byte COP instruction on the 65816 and a two byte NOP on the 65c02.

If I switch the emulator with a 65c02 emulator I wrote, then the code only works after this patch. 

The emulator is here https://github.com/differentprogramming/x16-emulator but the version you need isn't checked in.

mist64: > @mist64, that patch reveals another bug -- not related to this PR.
> The `status` byte isn't there anymore!

Thanks! Fixed in 70bfe30.

mist64: > Another place that I noticed is in Kernal's `MOUSE` subroutine. It steals zero-page location `$00` from us.

Thanks. Fixed in 1602060.

mist64:
I agree that `.byte $2c` is a problem generally, mostly because of the possible side effects of reading from random locations.

May I request though that you use the `bra :+` syntax in all cases that allow it? Until recently, I've tried to maintain the original coding style including its symbol name conventions, but I think we should use more modern conventions moving forward, and not introducing symbols in the "skip an instruction" case is I think a good new convention.

differentprogramming:
> I agree that `.byte $2c` is a problem generally, mostly because of the possible side effects of reading from random locations.
> 
> May I request though that you use the `bra :+` syntax in all cases that allow it? Until recently, I've tried to maintain the original coding style including its symbol name conventions, but I think we should use more modern conventions moving forward, and not introducing symbols in the "skip an instruction" case is I think a good new convention.

I'm not actually familiar with this assembler.  I noticed maybe 4 label conventions, I used whatever the convention around the given area was, since I don't know what is compatible with what.

But I **_THINK_** that if you have a named label you can't access a non-named label across it.

The instructions say that labels have colons but plenty of them don't.

Some are just :
Some are @number or @number: maye
or name: or just nameatstartofline

greg-king5:
https://cc65.github.io/doc/ca65.html#ss6.6
describes the kind of labels that Michael wants you to add in this PR (except where the target line already had a label).

differentprogramming:
Unnamed labels can not be referenced across named labels therefor there's only a couple labels I can change to be unnamed.

You can't have
jmp :++
aLable lda #5
: lda #6

greg-king5:
> Unnamed labels cannot be referenced across named labels; therefore, there's only a couple of labels I can change to be unnamed.
> 
> You can't have:
> ```asm
>         jmp :++
> aLable  lda #5
> :       lda #6
> ```

Your code example must use only one plus character:
```asm
        jmp :+
aLable  lda #5
:       lda #6
```
Then, it works.

differentprogramming:
I've never done this before.  The rom is a moving target.  Do I download a new version, merge the new changes then merge with my changes?

differentprogramming:
I've made the changes do I pull again?

mist64:
I manually merged this as 3d54229.

Thanks a lot!!

PR #89, opened 2019-11-19
Swap foreground/background PETSCII control code.

codewar65:
Swaps the foreground and background colors using ctrl-b ($02). Used for setting text bg color by setting fg color, swapping, and setting new fg color. Also can be used as a single code reverse toggle.

Untitled-1

### DIFF
--- a/kernal/editor.1.s
+++ b/kernal/editor.1.s
@@ -845,6 +845,18 @@ dnline	ldx tblx
 dwnbye	rts
 
 chkcol
+        cmp #$02    ; check ctrl-b for invert.
+        bne ntinv
+        lda color    ; get current text color.
+        asl a        ; swap msn/lsn.
+        adc #$80
+        rol a
+        asl a
+        adc #$80
+        rol a
+        sta color    ; stash back.
+        rts
+ntinv
 	ldx #15         ;there's 15 colors
 chk1a	cmp coltab,x
 	beq chk1b


mobluse:
I could download and update the ROM using Bash:
```
cd x16-rom
git fetch origin pull/89/head:patch-1
git checkout patch-1
make
cp rom.bin ../x16-emulator
```
@codewar65's example on https://murray2.com/threads/commander-x16-petscii-control-codes-question.438/page-2#post-3015 worked for me in x16emu in Raspbian Buster on Raspberry Pi 4 B.

In [aritm-x16.bas](https://github.com/commanderx16/x16-demo/blob/master/basic/aritm-x16.bas) for coming r35 I could change this line:
`1116 OPEN 1,0:POKE 713,0:PRINT CHR$($8E);CHR$($9E):SCREEN 0`
to:
`1116 OPEN 1,0:PRINT CHR$($8E);CHR$($90);CHR$($02);CHR$($9E):SCREEN 0`
and thus be independent of changing RAM addresses, and also be able to change background in [my filter program](https://github.com/mobluse/x16-petscii2utf8). The filter program, petscii2utf8, has not yet been updated to change background, i.e. handle Ctrl+B/$02.

There is already Reverse On/Reverse Off ($12/$92), so the importance of Ctrl+B is to change the background and also to be able to emit a character to a filter or serial port to tell that the background has changed.

codewar65:
> There is already Reverse On/Reverse Off ($12/$92), so the importance of Ctrl+B is to change the background and also to be able to emit a character to a filter or serial port to tell that the background has changed.

Accessing background text color via 'conio' is important to the artscene. As a format via petscii to share images.

PR #93, opened 2019-11-24
Added hexadecimal qualifier to immediate value in cbdos/sdcard.asm

MJoergen:
In the current commit, the line 253 is assembled to "d3f1 89 50    bit #$50". According to the comment in the source file, this is not what was intended.

Superficial testing shows no functional difference whether this patch is applied or not.

### DIFF
--- a/cbdos/sdcard.asm
+++ b/cbdos/sdcard.asm
@@ -250,7 +250,7 @@ sd_cmd_response_wait:
 @l:			dey
 			beq sd_block_cmd_timeout ; y already 0? then invalid response or timeout
 			jsr spi_r_byte
-			bitimm 80	; bit 7 clear
+			bitimm $80	; bit 7 clear
 			bne @l  ; no, next byte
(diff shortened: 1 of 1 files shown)


mist64:
Oh, I was aware that the original authors found this bug and fixed it, but I forgot to apply it here. They even wrote a blog post:

https://steckschwein.de/2019/10/16/weird-bug-in-sd-card-code/

It wasn't as easy as fixing the "$" though, iirc. I don't have the time to look right now, can you check?

greg-king5:
The `spi_r_byte` function returns the proper zero and sign flags. Therefore, that `bit` instruction and the `cmp` instruction after it aren't needed.  Just do a `bmi`:
```asm
jsr spi_r_byte
bmi @l  ; byte is invalid
rts
```

mist64:
The DOS/FAT/SD code has been replaced with a complete reimplementation in the meantime. Closing.

PR #114, opened 2019-12-26
Fix for mangled accumulator after color swap.

codewar65:
Possible reason for unwanted ISO swapping. May want an escape mechanism when returning from chkcol if color code / swap is handled so further control codes are not checked.


greg-king5:
The `chkcol` function has three exit paths.  The third path has the same problem as the first path; it needs a similar work-around (make it return `lda #$00`).

mist64:
I fixed both paths.

Issue #111, opened 2019-12-21
GRAPH: GRAPH_move_rect overlap support


mist64:
GRAPH_move_rect always copies the lines from top to bottom, which is incorrect if the two rectangles overlap and sy < ty.

A simple check should be made for this condition, and if it's true, a second code path should be used, which does the same as the original code path, but counts from the last line to the first line.

gaekwad:
I tried submitting a pull request for this.  Still learning 6502 assembly and it seemed like a straight-forward issue to look at.

mist64:
Fixed, thanks!

PR #116, opened 2019-12-28
fix GRAPH_move_rect overlapping bug

gaekwad:
Fix issue #111 with GRAPH_move_rect overlapping areas.

### DIFF
--- a/kernal/graph/graph.s
+++ b/kernal/graph/graph.s
@@ -508,13 +508,6 @@ GRAPH_draw_image:
 ;            r4   width
 ;            r5   height
 ;---------------------------------------------------------------
-_DecW:
-	lda 0,x
-	bne @1
-	dec 1,x
-@1:	dec 0,x
-	rts
-
 GRAPH_move_rect:
 	CmpW r3, r1
 	bcc @2
@@ -542,6 +535,13 @@ GRAPH_move_rect:
 	lda r5L
 	ora r5H
 	bne @2
+	rts
+
+_DecW:
+	lda 0,x
+	bne @1
+	dec 1,x
+@1:	dec 0,x
 	rts
 
 ;---------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


mac.inc
```diff
@@ -260,3 +260,11 @@
 	inc addr+1
 	@j:
 .endmacro
+
+.macro DecW addr
```
mist64: This is incorrect for e.g. the value of $00c0. $c0-1 is $bf, which is negative, so it would decrement the upper byte as well.

Look here for the correct implementation:
geos/kernal/math/dec.s

These macros were taken from GEOS. The reason GEOS didn't have DecW as a macro was probably because it will have to overwrite a register, which is unexpected, and because the zero-check is extra work, and it's not clear whether it's needed.

In fact, I'd inline the code and not create a macro.

kernal/graph/graph.s
```diff
@@ -509,18 +509,28 @@ GRAPH_draw_image:
 ;            r5   height
 ;---------------------------------------------------------------
 GRAPH_move_rect:
-; XXX overlaps
+	CmpW r1, r3
+	bpl @3
```
mist64: this might be correct, but bcc/bcs are more correct. :)

mac.inc
gaekwad: Ah. I didn't realize that the zero flag simply uses bit 7 as a sign bit.  I knew it couldn't be that easy. And of course, the values I tested with didn't happen to have bit 7 set in the lower byte. Also I was thinking, the screen is 320x200 right? So for the y coordinate and height, don't we not even have to worry about the upper byte and can just ignore it?

mist64: The current FB driver supports only 200px vertically, but the GRAPH_* API needs to support 16 bit coordinates, because the FB driver dan be replaced.

kernal/graph/graph.s
```diff
@@ -508,29 +508,40 @@ GRAPH_draw_image:
 ;            r4   width
 ;            r5   height
 ;---------------------------------------------------------------
+_DecW:
```
gaekwad: Is this what you meant by an inline function? I pretty much ended up copying Ddec.

mist64: Yes, copying. :-)

PR #120, opened 2019-12-29
readme-update

Esshahn:
added code block markdown to show invisible statement options


Esshahn:
already implemented in previous PR

PR #122, opened 2019-12-29
Charset fix2

Esshahn:
ã and Û


mist64:
Can you describe what you changed and why?

Esshahn:
I changed these chars: Û, î, ã
All three where different from their counterparts, e.g. having a thicker line.
They should now be more consistent.

PR #123, opened 2020-01-02
CLS statement

Esshahn:
simple shortcut for ?chr$(147)


basic/token2.s
```diff
@@ -58,6 +58,7 @@ gotk	=$cb
 
 ;**************************************
 reslst2	.byt "MO", 'N' + $80
+	.byt "CL", 'S' + $80
```
mist64: please add it at the end (before VPEEK) to keep the tokens more stable.

basic/x16additions.s
```diff
@@ -299,6 +299,11 @@ reset_copy:
 	stz d1prb 
 	jmp ($fffc)
 
+cls:
+	lda #$93
+	jsr $ffd2
```
mist64: jsr + rts = jmp

mist64: This should go through `outch`, which is the BASIC version of $FFD2 that adds error handling.

Also, the question is what should happen if the standard output is not the screen. Should we still just print this to the device? What does CLRSCR do in BASIC 7?

https://github.com/mist64/cbmsrc/tree/master/BASIC_C128

basic/token2.s
Esshahn: sure, thought it was alphabetically sorted

mist64: (At some point, we'll reshuffle them and see whether we can map some statement to BASIC 3.5/7/10 tokens for improved compatibility.)

Esshahn: done

basic/x16additions.s
Esshahn: `outch`-> done
`JSR + RTS`-> `JMP`(thanks ;) )

About the standard output. I haven't checked SCNCLR yet, maybe we should check what behavior is more useful?

* `SCNCLR` & `CLS` -> I expect the screen to be cleared regardless of the current output 
* `CLR` (not `CLS`) -> I expect the output to clear

Makes any sense?

greg-king5: > * `SCNCLR` & `CLS` -> I expect the screen to be cleared regardless of the current output
> * `CLR` (not `CLS`) -> I expect the output to clear

`CLR` erases a program's variables.

mist64: SCNCLR has to do a lot of extra work because of possible multiple windows, but in the end, it seems it just sends $93 to BSOUT.

PR #135, opened 2020-02-13
Fix CBDOS read_block to handle last block filling buffer completely

theelkmechanic:
Reading files from the SD card was failing if the file length was an exact multiple of the block size of 512 bytes. It turns out CBDOS read_block was doing a less than comparison instead of a less than or equals, so if the last block completely filled the buffer, it would not flag it internally as being the last block, so it would continue trying to read new blocks from the file without sending an EOF. Added a check to set the last block flag when the remaining bytes hits zero.


mist64:
Thanks!!

PR #137, opened 2020-03-01
Move fd_area after buffers to fix buffer alignment bugs

theelkmechanic:
Because fd_area was located before sd_blktarget/block_data, it shifted the base address of that buffer to $BC80. This was breaking fat_find_next and other directory iterations because they assume the buffer is aligned on a 256-byte boundary and only check the high byte when checking to see if they've reached the end of a directory block. This caused the last 4 of every group of 16 files on the SD card to be invisible. The proper fix would be to check both high and low bytes of the address, but if the buffer is properly aligned, it's faster to be able to just check the high byte. Moving fd_area after all the disk buffers fixes this issue. In future, we need to be careful not to insert anything before/in between the 512-byte buffers in cbdos/main.s that would change this alignment.

Proper fix in fat_find_next (similar fixes needed in other functions that use sd_blktarget or block_data):
```
@l6:
      lda dirptr+1
      cmp #>(sd_blktarget + sd_blocksize)       ; end of block reached?
      bcc ff_l4                 ; no, process entry
+      lda dirptr
+      cmp #<(sd_blktarget + sd_blocksize)
+      bcc ff_l4
```

### DIFF
--- a/cbdos/fat32.inc
+++ b/cbdos/fat32.inc
@@ -22,7 +22,7 @@
 ; SOFTWARE.
 
 .ifndef IMPORTED_FROM_MAIN
-.import fd_area, sd_blktarget
+.import fd_area, fd_area_end
 .endif
 
 ; TODO FIXME replace with errno.inc from cc65 api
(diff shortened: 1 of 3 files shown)


mist64:
The DOS/FAT/SD code has been replaced with a complete reimplementation in the meantime. Closing.

Issue #67, opened 2019-10-21
Optimizations for floating point arithmetic


MJoergen:
I have a number of ideas for optimizing the floating point arithmetic routines in the BASIC code.

For now I've forked this project and added the file [xadd.s](https://github.com/MJoergen/x16-rom/blob/master/basic/xadd.s), which contains an optimized version of the floating point addition. Additionally, I've added the file [test/add.bas](https://github.com/MJoergen/x16-rom/blob/master/test/add.bas) to do some rudimentary testing.

Benchmarking is difficult, because the execution time depends on the operand values. One data-point is the simple loop "FOR I=1 TO 10000; NEXT I", which now executes in 109 jiffies, as opposed to 116 jiffies in the original ROM. This corresponds to reducing each addition by - on average - 120 clock cycles.

I have (untested) ideas for optimization multiplication too, but I'm unsure how to proceed from here.

kktos:
Looks like a PR is in order ;)

MJoergen:
One thing we need to remember is that the original ROM has an entry point within the floating point addition routine, namely faddc, that is called from another place in the ROM. I believe it is used when converting a literal constant to floating point. So this needs to be handled. And then I guess we should remove the old code.

kktos:
Just replace the code with yours and we're set. I think @mist64 did what was needed to keep the fixed addresses fixed ;)

MJoergen:
I've just issued a Pull Request. I welcome any feedback.

MJoergen:
I've added a SQR implementation to my branch. It is more than five times faster than the original implementation.

The new implementation is more accurate too: During testing I noticed e.g. that the original ROM gave the value
SQR(1017) = 31.8904375
whereas the new implementation gives the value
SQR(1017) = 31.8904374
The correct value according to my calculator is close to 31.890437438.

Furthermore, the old ROM gave a non-zero value for SQR(9)-3, whereas the new implementation gives zero. In fact the new implementation gives exact integer results for all integer squares, up until SQR(139*139) which is not exactly 139.

PR #143, opened 2020-04-10
Optimized floating point routines: addition, multiplication, and SQR.

MJoergen:
This adresses issue #67 .

The speedup is approximately:
Addition : 10 %
Multiplication : 25 %
SQR : 80 %

In the directory test/ there are some benchmarks.

I've removed the floating point routine "mltply" from the jump table. It is not used anywhere in the ROM, and the optimized multiplication routine does not support it.


mist64:
This looks awesome! Do you want to add your name and a license to the new files?

In KERNAL, I use a comment on the top of the file like this:

```
;----------------------------------------------------------------------
; Console Library
;----------------------------------------------------------------------
; (C)2020 Michael Steil, License: 2-clause BSD

```

MJoergen:
@mist64 Yes, please add a license similar to the KERNAL, and include my name: Michael Jørgensen

mist64:
I just did a naive test of my own:

`A=1:FORI=1TO10000:A=A*1.0001:NEXT:?A`

The results with and without optimizations differ in the 8th digit after the decimal point.

Are the optimizations supposed to generate the exact same results?

MJoergen:
@mist64 Ideally yes, except that I fixed a bug in the original version of multiply:

If you try the following, then the original ROM will give two different results (one correct, the other incorrect):
A=16777217
PRINT 10*A
PRINT A*10

This I have fixed in my re-implementation of multiply, so this will certainly generate different results.

This is even more apparent in the SQR function, which will generate different (and more accurate) results than the ROM. For instance, SQR(9)-3 yields a non-zero result in the original ROM, but gives zero with my re-implementation.

In your naive test, what results did you get?

mist64:
```
before: 2.71814133
after:  2.71814132

```

mist64:
I merged it, and added a copyright/author/license header to all FPLIB source files in a separate commit.

This is very cool! Code from 1978 *significantly* optimized! :)

mist64:
Also, your patch fixes the C64 BASIC [Multiply Bug](https://www.c64-wiki.com/wiki/Multiply_bug)!

PR #147, opened 2020-05-13
CBDOS: Fix RAM bank 1 memory trashing when doing CBDOS detection

theelkmechanic:
The CBDOS detection vector calls sdcard_init directly with no wrapper to ensure it is using the correct bank 0. This causes it to trash 5 bytes at $BF9C in whichever bank happens to be selected, usually bank 1. This patch adds a wrapper to save and restore the bank and avoid the corruption.


mist64:
Excellent, thanks!

PR #148, opened 2020-05-20
Start at the right address when loading custom charset

RebeccaRGB:
Preserve X register in inicpy so cpycustom starts at the right address.


mist64:
Thanks!

PR #173, opened 2020-09-09
[DOS] Fix loading to the passed-in address when SA is 0

gaekwad:
The code was saving the old SA in X, which was getting clobbered before it tried to use it to determine whether to save using the disk address or passed-in address. This changes "talk" and "listn" to preserve X on the stack so they are not lost at the upper levels.

### DIFF
--- a/kernal/cbm/channel/load.s
+++ b/kernal/cbm/channel/load.s
@@ -56,8 +56,7 @@ ld20	cmp #4
 	lda #fndefault_end-fndefault
 	jsr setnam
 ;
-ld25	ldx sa          ;save sa in stack
-	phx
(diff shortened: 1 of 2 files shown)


kernal/cbm/channel/load.s
```diff
@@ -57,6 +57,7 @@ ld20	cmp #4
 	jsr setnam
 ;
 ld25	ldx sa          ;save sa in .x
```
greg-king5, line 59: The comment should say, "save sa on stack".

gaekwad: fixed

mist64:
Thank you very much! Sorry for the delay!

Issue #171, opened 2020-09-09
f4a5.load.s - STY vs STA


Zakero:
https://github.com/commanderx16/x16-rom/blob/9498437e4baf39b75a36b158f98f0ef9a59b4978/kernal/open-roms/iostack/f4a5.load.s#L33

The surrounding code makes me wonder if the "sty" should actually be "sta".

The intention looks like MEMUSS is being copied to STAL.  In kernelstatus_reset, the A register is set to #$00 so "MEMUSS+1" is not being used there.  Which leads me to believe that line 33 should be "sta" but I do not fully understand the details and could be wrong.

greg-king5:
You aren't wrong.  It should be a simple 16-bit (two-byte) copy from `MEMUSS` to `STAL`.

PR #174, opened 2020-09-09
Bug Fix: The wrong register was loaded into STAL+1

Zakero:
This pull request is to correct the issue discussed in [f4a5.load.s - STY vs STA](https://github.com/commanderx16/x16-rom/issues/171#issue-696453552).


mist64:
I submitted this upstream:

https://github.com/MEGA65/open-roms/issues/66

Issue #49, opened 2019-10-04
HEX$(), BIN$() as counterparts to $A000 and %11001100


tomxp411:
As suggested here:

https://murray2.com/threads/suggestion-for-additional-basic-commands.328/post-2010

We have the $ prefix for hexadecimal numbers and the % prefix for binary numbers. We should have HEX$() and BIN$() functions to print numbers out in hex or binary format.

mist64:
Sounds great. Should be easy. :)

JimmyDansbo:
I have been trying to create the HEX$() function, but I can not seem to get my head around how to actually add the command to BASIC.
So far, this is what I have done:
* Created function in x16additions.s, called hexd (I see that chr$, str$ and so on are named like that)
* Added '.word hexd' in tokens.s in the stmdsp2 list under '.word joy'
* Added '.byt "HEX", '$' + $80' in token2.s in the reslst2 list under '"byt "JO", 'Y' + $80'
* Changed num_esc_functions from 5 to 6

Still it seems that BASIC does not recognize the new keyword, can someone point me in the direction of what I am missing?

JimmyDansbo:
So, I managed to get HEX$ recognized as a command/token.
Instead of .byt "HEX", '$' + $80 I had to do .byt "HEX", $a4
It seems that '$'+$80 is not correct, probably because ascii value of '$' is not the same as the petscii value.

Now I need help figuring out how to return a string from my function. Looking at existing functions have not gotten me on the correct track yet. :(

greg-king5:
> Instead of `.byt "HEX", '$' + $80`, I had to do `.byt "HEX", $a4`.
> It seems that `'$'+$80` is not correct, probably because ascii value of '$' is not the same as the petscii value.

That is strange.  `'$'+$80` is `$a4`, in both ASCII and PetSCII.

PR #176, opened 2020-10-24
Implementation of BIN$ & HEX$ functions

JimmyDansbo:
An implementation of the BIN$ and HEX$ functions as suggested in issue #49 
Both functions only work with byte input because of space restraints in the 16KB BASIC ROM page.
The hexd function does a jsrfar to the byte_to_hex_ascii function in monitor.s, but the address is hard coded as symbols are not shared between BASIC and MONITOR pages.

### DIFF
--- a/basic/x16additions.s
+++ b/basic/x16additions.s
@@ -54,12 +54,14 @@ monitor:
 	jsr bjsrfar
 	.word $c000
 	.byte BANK_MONITOR
+	; does not return
 
 ;***************
 geos:
 	jsr bjsrfar
 	.word $c000 ; entry
 	.byte BANK_GEOS
+	; does not return
 
 ;***************
 color	jsr getcol ; fg
--- a/basic/x16additions.s
+++ b/basic/x16additions.s
@@ -92,25 +92,48 @@ coltab	;this is an unavoidable duplicate from KERNAL
 
 ; convert byte to binary in zero terminated string and
 ; return it to BASIC
-bind	jsr chrget ; get char
+bind:	jsr chrget ; get char
 	jsr chkopn ; check opening paren
-	jsr getbyt ; byte to convert
+	jsr frmadr ; get 16 bit word in Y/A
 ; lofbuf starts at address $00FF and continues into
 ; page 1 ($0100).
 ; Forcing address size to 16bit allows us to cross
 ; from zero page to page $01 when using .X as an index
-	txa
-	ldx #7
-@loop:	lsr        ; low bit to carry
-	tay	   ; save .A for next iteration
+	ldx #0
+	phy	   ; save low byte
+	tay
+	cmp #0
+	beq @LOWBYTE
+@HIGHLOOP:
+	asl	   ; high bit to carry
+	pha	   ; save .A for next iteration
 	lda #'1'
 	bcs :+
 	dec
-:	sta a:lofbuf,x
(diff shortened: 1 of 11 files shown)


inc/mac.inc
```diff
@@ -20,7 +20,7 @@
 		lda #<(value)
 		sta dest
 	.endif
-	.if .const(value) .and value=0
+	.if (.const(value) .and .hibyte(value)=0)
```
greg-king5, line 23: You can use that same improvement for the low byte of `LoadW`.

JimmyDansbo: You are right, I missed that :-)

basic/x16additions.s
```diff
@@ -90,6 +90,71 @@ coltab	;this is an unavoidable duplicate from KERNAL
...
+; lofbuf+1 = $0100 so .X can be used as index
+; without trying to cross page
+	txa
+	ldx #7
+@loop:	lsr        ; low bit to carry
+	tay	   ; save .A for next iteration
+	lda #'1'
+	bcs :+
+	dec
+:	sta lofbuf+1,x
```
mist64, line 111: you can use
   sta a:lobuf,x
to force
  STA $00FF,X
which will not wrap around in zero page.

mist64:
Sorry it took to song for me to look at this.

I like the feature, and I like the code, but I really think we should support 16 bit values, like on the C128.

By the way, we can legally use the C128 source:

https://github.com/mist64/cbmsrc/blob/master/BASIC_C128/hexfunc.src

JimmyDansbo:
Nice to see you back, Michael :)
I am not sure when I will have time to look into this further. Should I just close this pull request?
Hopefully someone else can help with implementing 16bit version of these functions.

mist64:
Leave it open. I'll think of something.

basic/x16additions.s
JimmyDansbo: I was not aware of the ability to force the address size of a symbol in ca65, but found it in the documentation here: https://cc65.github.io/doc/ca65.html#ss8.3 - Thanks.

inc/mac.inc
```diff
@@ -14,13 +14,13 @@
 .endmacro
 
 .macro LoadW dest, value
-	.if .const(value) .and value=0
+	.if .const(value) .and .lobyte(value)=0
```
mist64, line 17: 👍

basic/x16additions.s
JimmyDansbo: Now fixed

inc/mac.inc
JimmyDansbo: :smile:

JimmyDansbo:
@mist64 , would you prefer the HEX$ and BIN$ functions to precede the returned string with '$' and '%' respectively? At the moment they are returning strings with varying length depending on the number of bits being passed to them (8 or 16).

mist64:
The C128 always returns 4 digits for HEX$() without a prefix and does not support BIN$(). I'm tempted to say go with C128 compatibility...

But... I won't say that. Variable length is better. But no prefix.

JimmyDansbo:
Great, as far as I can see, that should be the current functionality.
Unless I have f*cked something up, I hope this can be merged?

mist64:
:)

irmen:
awesome to see this merged with support for words

PR #180, opened 2020-12-10
fix crash in FB_set_pixels for count>255. Fixes #179

### LINKED ISSUE #179: FB_set_pixels crashes when count > 255
FB_set_pixels with a count of 256 or more crashes.
I think it's caused by the jsr that it's doing internally and that it is not properly dealing with the double loop invocation or some thing.

What confuses me is that FB_set_pixels is defined twice:

https://github.com/commanderx16/x16-rom/blob/a853afd4ff164f4b9b63762bf6e9f1931e6683f8/kernal/drivers/x16/framebuffer.s#L176
and also here
https://github.com/commanderx16/x16-rom/blob/31a9470eb75ae2e5c0fd3ff9617a78491e7df024/kernal/graph/drv_ram.s#L164

irmen:
FB_set_pixels treated the restore of the registers wrong so it crashed when count is > 255. This has been corrected. Fixes #179 
Notice that FB_get_pixels already worked correctly.

### DIFF
--- a/kernal/drivers/x16/framebuffer.s
+++ b/kernal/drivers/x16/framebuffer.s
@@ -176,7 +176,12 @@ FB_get_pixel:
 FB_set_pixels:
 	PushB r0H
 	PushB r1H
+	jsr set_pixels_FG
+	PopB r1H
+	PopB r0H
+	rts
 
+set_pixels_FG:
 	lda r1H
 	beq @a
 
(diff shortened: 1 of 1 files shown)


irmen:
it's a tiny fix only 7 lines

mist64:
Thanks! Sorry for the delay...

irmen:
thanks for merge

PR #187, opened 2021-01-11
Custom key listener

stefan-b-jakobsson:
This is a pull request illustrating a solution to an issue I opened earlier.

Essentially it does the following. At the end of the function receive_scancode in kernal/drivers/x16/ps2kbd.s it makes an indirect jump to the address stored in $02BB-BC.

This makes it possible to intercept a PS/2 scan code before it's handled by the Kernal's default keyboard routines. You may use this to get finer control over modifier keys (my initial interest). But you may also use it to catch key up and release events, which might be of interest if you for instance are programming keyboard controlled games. You may pretty much change every aspect of how the Kernal reacts to the keyboard.

A small test program intercepting the Ctrl key, printing the letter D on key down and the letter U on key up. To init the keylistener, call the setup code.

```
setup:
sei
lda #<keylistener
sta$02bb
lda #>keylistener
sta $02bc
cli
rts

keylistener:
php   ;Store these values on the stack, need to be restored if not Ctrl key
pha
phx

bcs keyup   ;receive_scancode returns C=1 if keyup event

keydown:
cmp #$14   ;PS/2 scan code for left Ctrl key
bne not_ctrl
lda #$44
jsr $ffd2    ;Kernal print routine
bra ctrl_exit

keyup:
cmp #$14
bne not_ctrl
lda #$55
jsr $ffd2

ctrl_exit:
plx     ;Pull values we stored at beginning off the stack
pla
plp
lda #0    ;By setting A=0 we will consume this key event
tax         ;X=scan code prefix, maybe it mustn't be set to zero
rts         ;Return control to the Kernal

not_ctrl:
plx         ;Restoring X, A and C from stack
pla
plp
rts         ;Return control to the Kernal, handles the key normally

```

### DIFF
--- a/cfg/kernal-x16.cfgtpl
+++ b/cfg/kernal-x16.cfgtpl
@@ -16,6 +16,8 @@ SEGMENTS {
 	VARFONTS:   load = KVAR,     type = bss;
 	KVAR2:      load = KVAR2,    type = bss;
 
+	KEYEVT:		load = KEYEVT,	 type = bss;
+
 	GDRVVEC:    load = GDRVVEC,  type = bss;
 	KVECTORS:   load = KVECTORS, type = bss;
 	KVARSB0:    load = KVARSB0,  type = bss, define=yes;
--- a/cfg/x16.cfginc
+++ b/cfg/x16.cfginc
@@ -21,6 +21,7 @@ ZPBASIC:  start = $00D4, size = $002B; # BASIC (last byte used: $FE)
 
 /* $0200-$02FF: always-available variables and RAM code */
 KVAR:     start = $0200, size = $00BB; # KERNAL
+KEYEVT:   start = $02BB, size = $0002; # KERNAL KEY EVENT HANDLER JUMP VECTOR
 /*        start = $02BB, size = $0009; # reserved for KERNAL growth */
 KERNRAM:  start = $02C4, size = $0020; # KERNAL RAM code
 GDRVVEC:  start = $02E4, size = $001C; # framebuffer driver vectors
--- a/kernal/cbm/memory.s
+++ b/kernal/cbm/memory.s
@@ -8,15 +8,19 @@
 
 .include "io.inc"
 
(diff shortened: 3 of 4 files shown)


mist64:
Excellent!

But there's one issue I have with this: I'd like to get away from magic memory addresses. It's common to consider all memory addresses on the C64 as API, and the Programmer's Reference Guide kind of documented most of them, but they are inherently incompatible between Commodore-like machines.

There is one address range though that is documented and mostly stable across machines: the "KVECTORS" segment at $0300-$0333. The C64 even has 8 bytes reserved at the end of it.

I'd suggest extending KVECTORS by 2 bytes and moving KVAR2 and KERNRAM2 up. This fills the 2 bytes "reserved for KERNAL growth" at $03CA-03CB.

bobbyjim:
So that means the "keyboard/keyscan handler" address would be stored at $03CA-B?

mist64:
No no, the new vector should be the 27th vector in the list of vectors at $0300.

The  vectors used to be $0300-$0333; they should now be $0300-$0335, and the new vector should be at $0334/$0335.

The data that used to start at $0334 needs to be moved up.

bobbyjim:
A hook for a monitor seems really nice to me... but I think by "different" and "don't currently support", I think you mean the monitor we currently have is not the built-in PET monitor, and anyway probably WON'T be the build-in PET monitor, so.... well there's no reason to preserve that vector.

OK, I'll buy that.

mist64:
Yes, the monitor we have is not the PET/CBM/TED/C128 monitor, for two reasons:
* When we started, we didn't have the source to the monitor.
* Our monitor has more features already anyway. :)

stefan-b-jakobsson:
I tested what @mist64 proposed. Seems to be working.

Since I'm a git noob I have to look a bit more on that aspect before updating.

stefan-b-jakobsson:
I made a new pull request from a new branch after merging in the changes made to the master. Therefore closing this.

PR #188, opened 2021-01-16
clarified Position command in CMD-DOS

tomxp411:
I just added some text to describe how to use the new "P" command in CMD-DOS. 

I'll add a database demo separately over on the forum.


greg-king5:
The channel number, also, is binary.  Channel eight should be `+chr$(8)+`.  Only the `"p"` is PetSCII.

tomxp411:
What I wrote is how it’s working right now in the emulator. When I sent CHR$(8), I got a blinking light, and a NO CHANNEL error. 

I will try it again when I get home tonight, but the reason I wrote the clarification was because the only way I could get it to work was with the PETSCII 8 character.

tomxp411:
Ok, I figured out is that either will work, so I'm guessing CMDR DOS is just taking the bottom 4 bits. I don't know why it didn't work for me the first time I tried with CHR$(8), but stuff happens. :)  I'll fix it.

PR #198, opened 2021-04-06
Modify screen $80 mode to be 320x240. 

gaekwad:
Also moves character tile data to $04000 to have a larger block of contiguous free space.

### DIFF
--- a/basic/graph.s
+++ b/basic/graph.s
@@ -102,10 +102,10 @@ get_point:
 	lda poker
 	sta y1L
 	sec
-	sbc #<200
+	sbc #<240
 	lda poker+1
 	sta y1H
-	sbc #>200
+	sbc #>240
 	bcs linfc
 	rts
 
@@ -142,10 +142,10 @@ get_points_col:
 	lda poker
 	sta y2L
 	sec
-	sbc #<200
+	sbc #<240
 	lda poker+1
 	sta y2H
-	sbc #>200
+	sbc #>240
 	bcs linfc
 
 	jsr get_col
--- a/basic/test.s
+++ b/basic/test.s
@@ -75,7 +75,7 @@ test2_vline:
 	LoadW r0, 1
 	LoadW r1, 6
 	LoadW r2, 1
-	LoadW r3, 198
+	LoadW r3, 238
 	lda #0 ; set
 	jsr GRAPH_draw_line
 
@@ -83,7 +83,7 @@ test2_vline:
 	lda #4
 	jsr GRAPH_set_colors
 	LoadW r0, 3
-	LoadW r1, 198
+	LoadW r1, 238
 	LoadW r2, 3
 	LoadW r3, 6
 	lda #0 ; set
@@ -162,9 +162,9 @@ test4_set_get_pixels:
 	; print result of comparison
 	lda r1H
 	bne @2
-	LoadW 0, str_BAD
+	LoadW r15, str_BAD
 	bra @3
-@2:	LoadW 0, str_OK
+@2:	LoadW r15, str_OK
 @3:	lda #9
 	jsr GRAPH_set_colors
 	LoadW r0, 263
@@ -215,9 +215,9 @@ test5_filter_pixels:
 	; print result of comparison
 	lda r1H
 	bne @2a
-	LoadW 0, str_BAD
+	LoadW r15, str_BAD
 	bra @3a
-@2a:	LoadW 0, str_OK
+@2a:	LoadW r15, str_OK
 @3a:	lda #10
 	jsr GRAPH_set_colors
 	LoadW r0, 263
@@ -428,7 +428,7 @@ test12_char_styles:
 @1:	inx
 	cpx #5
 	bne @2
-	LoadW 0, test_string
+	LoadW r15, test_string
 	jsr print_string
 	ply
 	iny
@@ -514,7 +514,7 @@ checksum_framebuffer:
 	
 	plx
 	inx
-	cpx #200
+	cpx #240
 	bne @loop
 
 	lda #0
@@ -522,7 +522,7 @@ checksum_framebuffer:
 	jsr GRAPH_set_colors
 
 	LoadW r0, 295
-	LoadW r1, 190
+	LoadW r1, 230
 	LoadW r2, 24
 	LoadW r3, 9
 	jsr GRAPH_set_window
@@ -570,8 +570,8 @@ hextab:
 	.byte "0123456789ABCDEF"
 
 ; http://www.6502.org/source/integers/crc-more.html
-crclo	=0              ; current value of CRC
-crchi	=1              ; not necessarily contiguous
+crclo	=r7         ; current value of CRC
+crchi	=r7+1       ; not necessarily contiguous
 
 crc16_f:
 	eor crchi       ; A contained the data
@@ -606,7 +606,7 @@ crc16_f:
 
 print_string:
 	ldy #0
-:	lda (0),y
+:	lda (r15),y
 	beq :+
 	phy
 	jsr GRAPH_put_char
@@ -669,7 +669,7 @@ test2:
 	jsr GRAPH_draw_line
 	IncW r1
 	IncW r3
-	CmpWI r1, 200
+	CmpWI r1, 240
 	bne :-
 	
 	lda #0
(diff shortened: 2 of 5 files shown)


gaekwad:
Nevermind, this is acting pretty buggy. That's what I get for only doing some testing. :)

mist64:
It's more complicated than this, because all graphics code needs to take care of the 17th bit when calculating addresses.

gaekwad:
> 
> 
> It's more complicated than this, because all graphics code needs to take care of the 17th bit when calculating addresses.

Yup! Last night I started going through to fix each such place where it's being ignored/is hardcoded.  I should hopefully be done today if I'm not too busy.

gaekwad:
I think I fixed the issues as far as I can tell.  At the least, the "TEST", "TEST2", and "TEST3" commands seem to work fine.

greg-king5:
It seems to me that you're going in the wrong direction!  The font should be moved up, not down.

It should be put as high in the VRAM address space as possible.  Then, the sprites should be put below that.  Then, between the sprites and the text screen should be a _single_ "open" block of VRAM.  (There should be NO other blocks of "-- unused --" VRAM!)  Various geometries of bitmaps can be put in that single open area.  When sprites aren't used, their area can be included in the single open block.

That would be the most efficient use of VRAM.

gaekwad:
> 
> 
> It seems to me that you're going in the wrong direction! The font should be moved up, not down.
> 
> It should be put as high in the VRAM address space as possible. Then, the sprites should be put below that. Then, between the sprites and the text screen should be a _single_ "open" block of VRAM. (There should be NO other blocks of "-- unused --" VRAM!) Various geometries of bitmaps can be put in that single open area. When sprites aren't used, their area can be included in the single open block.
> 
> That would be the most efficient use of VRAM.

I agree that you want as big of a block of free space as possible.  However, I'd argue that, when not using $80 screen mode, it doesn't matter whether the character data is at $4000 or $1F1C0, the size of free space is the same.  When you are using screen $80 mode, most of that free space is being used by the 320x240 bitmap anyway, and you normally wouldn't be concerned with the free space when using it.  

If you're looking at the comment in inc/io.inc you can ignore that unused space from $3C00-$4000. I was going on the assumption that the comment was correct that "text mode" (layer 1 character map) goes from $0000-$3BFF, but I believe this is incorrect, the "text mode" goes from $0000-$3FFF.  I should probably make a commit to fix the comment regarding this.

You'll notice, then, that the only other "unused" block is $1F400-$1F9BF, but this block is only there when using screen $80 mode, otherwise that would be combined with all the space the 320x240 bitmap takes up.

greg-king5:
> If you're looking at the comment in "inc/io.inc", you can ignore that unused space from $3C00-$3FFF. I was going on the assumption that the comment was correct that "text mode" (layer 1 character map) goes from $0000-$3BFF, but I believe this is incorrect, the "text mode" goes from $0000-$3FFF. I should probably make a commit to fix the comment regarding this.

Sure, the tile map reaches up that far, but only 60 of those rows are displayed on the standard text screen.  Is it impossible to use the rest of that VRAM area to hold something else, such as sprite data?

Elektron72:
> Is it impossible to use the rest of that VRAM area to hold something else, such as sprite data?

It should be possible to use the last four rows of the tilemap to store other data, as long as the layer is never scrolled down. However, the bigger issue with putting the character data at $4000 is that it assumes that the default screen configuration is being used. If someone wanted to use the standard character set with a smaller tilemap, it would split VRAM; if someone wanted to use the standard character set with a larger tilemap, it would be overwritten. The only thing that we can be sure of is that VRAM ends at $1F9BF. Even if it creates a small block of unused space, I believe that near the top of VRAM would be the most efficient place for the character set, as it makes no assumptions about how the rest of VRAM is used.

gaekwad:
> 
> 
> > If you're looking at the comment in "inc/io.inc", you can ignore that unused space from $3C00-$3FFF. I was going on the assumption that the comment was correct that "text mode" (layer 1 character map) goes from $0000-$3BFF, but I believe this is incorrect, the "text mode" goes from $0000-$3FFF. I should probably make a commit to fix the comment regarding this.
> 
> Sure, the tile map reaches up that far, but only 60 of those rows are displayed on the standard text screen. Is it impossible to use the rest of that VRAM area to hold something else, such as sprite data?

Ah, I see why it's marked as unused now. I forgot about the fact that there are extra rows of tiles due to the limitation of the selection of row/height of the tilemap.

> 
> 
> > Is it impossible to use the rest of that VRAM area to hold something else, such as sprite data?
> 
> It should be possible to use the last four rows of the tilemap to store other data, as long as the layer is never scrolled down. However, the bigger issue with putting the character data at $4000 is that it assumes that the default screen configuration is being used. If someone wanted to use the standard character set with a smaller tilemap, it would split VRAM; if someone wanted to use the standard character set with a larger tilemap, it would be overwritten. The only thing that we can be sure of is that VRAM ends at $1F9BF. Even if it creates a small block of unused space, I believe that near the top of VRAM would be the most efficient place for the character set, as it makes no assumptions about how the rest of VRAM is used.

True.  I liked the idea of keeping the layer 1 tilemap/data together, but if we're worried about the tilemap being different sizes then they can't be.

gaekwad:
So this is how the memory layout looks after I do the calculations for moving the character tile data to the top of VRAM.  Unfortunately, due to the limitation of only specifying the first 6 bits of the tile data address, this actually results in more small pockets of free space.  Also, the $80 mode 320x240 bitmap needed to be moved lower in memory, and due to the top 6-bit limitation, it ends up being at $4000, but this only matters in $80 mode.  Outside of that, $03C00-$16FFF are free. And, unless you're utilizing those sprite functions, you could ignore those sprite locations as well, in which case $03C00-$1EFFF are free.

```
 ;
 ; VERA Video RAM Layout
 ;
 ; $0:0000 - $0:3BFF  Text Mode
 ; $0:3C00 - $0:3FFF  -- unused --
 ; $0:4000 - $1:6BFF  320x240@256c Bitmap
 ; $1:6C00 - $1:6FFF  -- unused --
 ; $1:7000 - $1:EFFF  Sprites
 ;                    $1:7000  Sprite #0
 ;                    $1:8000  Sprite #1
 ;                    $1:9000  Sprite #2
 ;                    $1:A000  Sprite #3
 ;                    $1:B000  Sprite #4
 ;                    $1:C000  Sprite #5
 ;                    $1:D000  Sprite #6
 ;                    $1:E000  Sprite #7
 ; $1:F000 - $1:F7FF  Charset
 ; $1:F800 - $1:F9BF  -- unused --
 ; $1:F9C0 - $1:FFFF  VERA registers
```

Another option would be to move the beginning of the sprite location to $3C00 as discussed.  However, permanently allocating the unused part of the tilemap for the sprites makes me nervous.  What if the user wanted to be able to scroll it?  Especially considering that we're taking in to account if the user wanted to make a larger tilemap, it seems strange to ignore that possibility.

Elektron72:
I've never really used the KERNAL graphics API (which relies on a lot of these addresses), so feel free to make any optimizations you want. As I believe that many applications will ignore the API and write directly to VERA, my main goal is to ensure that the largest possible contiguous block of VRAM can be used without overwriting the character set (although putting the character set at $00000 would likely achieve this goal more efficiently, moving the text mode tilemap would break a lot of code, so it would need to be discussed with the team and the community). Flexibility is not as important to me if the KERNAL API is being used, so making use of the block beginning at $03C00 is an option.

gaekwad:
> 
> 
> I've never really used the KERNAL graphics API (which relies on a lot of these addresses), so feel free to make any optimizations you want. As I believe that many applications will ignore the API and write directly to VERA, my main goal is to ensure that the largest possible contiguous block of VRAM can be used without overwriting the character set (although putting the character set at $00000 would likely achieve this goal more efficiently, moving the text mode tilemap would break a lot of code, so it would need to be discussed with the team and the community). Flexibility is not as important to me if the KERNAL API is being used, so making use of the block beginning at $03C00 is an option.

Well, that's the thing, the sprite location stuff is being utilized by some sprite functions that the kernal uses in some places (such as the mouse sprite).  It's not specific to screen $80 mode, which is why I'm nervous to set it to $3C00.  The reason I say that in most cases you can ignore the sprite stuff is because, I assume if you're concerned with the VRAM's free space and whatnot, you're probably writing a big game or graphics program that takes over the system anyway, so you don't have to worry about the kernal's use of those functions. (Unless your game make's use of the kernal's mouse routines maybe, or you want to use sprite_set_image and sprite_set_position yourself).

Anyhow, if there's no objection, I'll commit the changes as shown above.

mist64:
I took this code, picked a different memory layout ;-) (FB at 0 – makes math easier) and fixed the bit 16 carry. It's been merged here: 3ff8eb2462e83156128f8176ebeccdad3c6637d5

Issue #200, opened 2021-04-08
recent change made joysticks no longer work


irmen:
The changes made in this commit a few days ago d2b137bf3a1d2c5afe57f25346f007986a502c05 killed the joysticks.
Running the new emulator with ROMs built without this change, still work.

test program in BASIC i'm using:

10?JOY(1):GOTO 10

the values don't change anymore since this commit above.
(SDL still picks up the controller as shown by the debug messages printed on the console)

indigodarkwolf:
Looks like the return value in the accumulator for joystick_get() is getting lost because the macro specifically stores it at the start of the function and then retrieves it at the end of the function. So that needs to change to only preserve the value of the accumulator at the end of the function. 

Possibly replace the macro at the start with `KVARS_START_TRASH_A_NZ`, then change the end of the function to stow A to kvswitch_tmp1, do `KVARS_END_TRASH_A_NZ`, then retrieve A from kvswitch_tmp1 before the `rts`?

indigodarkwolf:
Ah, well, I had the right idea, but the fix had to be slightly different. PR incoming.

irmen:
above PR was merged and solved this issue.

PR #201, opened 2021-04-09
Fix issue #200

indigodarkwolf:
KVARS_START_TRASH_NZ and KVARS_END_TRASH_NZ are preserving A across the entire function call, in the one place where they're used this seems to be both unnecessary and incorrect, the A only needs to be preserved across the macro, and explicitly *should not* be preserved across the whole call. This seems to fix joystick polling, verified with the following test:

10?JOY(1):GOTO10
RUN

See also issue #200

### DIFF
--- a/inc/banks.inc
+++ b/inc/banks.inc
@@ -90,9 +90,11 @@ imparm     = $E2
 	lda ram_bank
 	pha
 	stz ram_bank
+	lda kvswitch_tmp1
 .endmacro
 
 .macro KVARS_END_TRASH_NZ
+	sta kvswitch_tmp1
 	pla
 	sta ram_bank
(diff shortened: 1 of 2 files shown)


greg-king5:
`KVARS_START_TRASH_X_NZ` should be used at the beginning of `joystick_get:`.

indigodarkwolf:
Yeah, that seems fair. I still think that each macro was only intended to preserve the value of A for the duration of the macro and not the entire function, but it's a little less code and a little faster to use KVARS_START_TRASH_X_NZ in any event.

indigodarkwolf:
Done and done, and tested with the same BASIC as above.

greg-king5:
I agree with you about those macro fixes.

PR #202, opened 2021-04-12
Integrate CodeX Interactive Assembly Environment into ROM

mjallison42:
First integration of the CodeX Interactive Assembly Environment into ROM. There are likely to be future bug fixes, etc.


basic/token2.s
```diff
@@ -58,6 +58,7 @@ gotk	=$cb
 
 ;**************************************
 reslst2	.byt "MO", 'N' + $80
+	.byt "C", 'X' + $80
```
mist64, line 61: can you move the new keyword to the end of the list, so the tokens are not renumbered? otherwise, this would break all BASIC programs SAVEd.

mist64 (changes requested):
Great. Just one issue; and consider changing to TABs instead of spaces (and no TAB/spaces between mnemonic and operand?) for consistency with the majority of the source in this repo. But as mentioned to you privately before, this is entirely your choice, since it's a pretty much standalone sub-project. I just suggest it so future edits by different people are less likely to completely mess up the formatting.

basic/token2.s
mjallison42: e.g. move after BIN$ ?

tomxp411:
This seems to be failing on the R39 emulator, as per 

https://www.commanderx16.com/forum/index.php?/topic/2585-codex/

mjallison42:
I will be submitting a PR next week with a series of fixes.

Mike A

PR #204, opened 2021-04-16
Fix a race condition in joystick_get

### LINKED ISSUE #203: joystick_get() and JOY() are corruptable by irq handler in v39 [includes patch]
I discovered that there is a race condition in the joystick_get() kernal routine (and thus also in the basic JOY() function): it seems that when it gets interrupted by the system's irq handler, the values returned from the routine can be corrupted, thus sending fake joystick triggers to the program.

Here is an attempt to reproduce the problem in a basic program:
image

The result shown is the program running for about a minute while not touching the joypad or keyboard.

This is a tricky issue that we found is very timing sensitive.  Sometimes it takes a LONG time before the first ERROR appears. Sometimes you need to tweak the program a bit for instance by introducing a variable.  Enabling WARP mode speeds up the process.   However in my own program that I work on, the issue almost always immediately and consistently pops up at a certain point int he program, so it is a real problem (for me at least).  

The big difference we think from my code vs other people's code, is that I am not reading the joystick status from within my own irq handler but simply from the normal program flow (and system irq handler is still running normally) -- like the basic program above does too.  That may explain the fact that nobody else ran into this yet.  (assumption)

As far as I was able to test, the issue does not occur on the previous V38 version of the emulator + roms.

A possible fix revolves around disabling IRQ during the joystick_get(). Suggestion:
- at least make the basic JOY() function do a SEI / CLI to avoid returning corrupted results
- joystick_get() can't do this because then you can't use it anymore from within an IRQ handler
- joystick_get() could perhaps check processor status bit to see if it is running inside IRQ handler and if not, do the SEI/CLI. 
- if joystick_get() doesn't get changed to make it robust against this, at least the documentation should mention that user code should SEI/CLI if not calling from an irq handler
- or perhaps whatever it was that got changed from v38 to v39 can be changed back...

Elektron72:
KVARS_START_TRASH_NZ and KVARS_END_TRASH_NZ are not interrupt-safe. If the macros are interrupted by an ISR that also uses the KVARS macros, kvswitch_tmp1 will likely be overwritten. This pull request modifies joystick_get to use KVARS_START and KVARS_END, which are interrupt-safe. This change fixes #203. This pull request also adds a warning about interrupt safety to inc/banks.inc.

### DIFF
--- a/inc/banks.inc
+++ b/inc/banks.inc
@@ -84,6 +84,7 @@ imparm     = $E2
 	stx ram_bank
 .endmacro
 
+; Warning: KVARS_START_TRASH_NZ and KVARS_END_TRASH_NZ are not interrupt-safe
 .macro KVARS_START_TRASH_NZ
 .import kvswitch_tmp1, kvswitch_tmp2
 	sta kvswitch_tmp1
--- a/inc/banks.inc
+++ b/inc/banks.inc
@@ -84,7 +84,8 @@ imparm     = $E2
 	stx ram_bank
 .endmacro
 
-; Warning: KVARS_START_TRASH_NZ and KVARS_END_TRASH_NZ are not interrupt-safe
+; Warning: KVARS_START_TRASH_NZ and KVARS_END_TRASH_NZ can cause a race
+; condition if interrupts are enabled.
 .macro KVARS_START_TRASH_NZ
 .import kvswitch_tmp1, kvswitch_tmp2
 	sta kvswitch_tmp1
--- a/kernal/drivers/x16/joystick.s
+++ b/kernal/drivers/x16/joystick.s
@@ -106,7 +106,7 @@ l1:	lda nes_data
 ;              on NES and SNES.
 ;---------------------------------------------------------------
 joystick_get:
-	KVARS_START_TRASH_X_NZ
+	KVARS_START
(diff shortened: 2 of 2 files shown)


kernal/drivers/x16/joystick.s
```diff
@@ -106,7 +106,7 @@ l1:	lda nes_data
 ;              on NES and SNES.
 ;---------------------------------------------------------------
 joystick_get:
-	KVARS_START_TRASH_X_NZ
+	KVARS_START
```
greg-king5, line 109: That change isn't needed!  `KVARS_START_TRASH_X_NZ` _is_ interrupt-safe.  And, all of the start macros balance all of the end macros (only the old RAM bank number is pushed/popped across subroutines).

inc/banks.inc
```diff
@@ -84,6 +84,7 @@ imparm     = $E2
 	stx ram_bank
 .endmacro
 
+; Warning: KVARS_START_TRASH_NZ and KVARS_END_TRASH_NZ are not interrupt-safe
```
greg-king5, line 87: That wording is misleading.  It implies that they can't be used in interrupt handlers.  But, those macros are dangerous only where interrupts are enabled.

PR #207, opened 2021-04-17
Remove extra space from BASIC error messages

### LINKED ISSUE #175: Don't replicate C64 error message bugs
The C-64 version of 2.0 BASIC was somewhat quickly ported from the VIC-20 version and one of the artifacts is an extra space between the error type and the word ERROR - that space was a carriage return on the VIC. This was corrected to a single space in Commodore BASIC 3.5 and 7.0; it would be nice not to be quite so faithful to C64's particular frozen version. :)

Elektron72:
Fixes #175.

### DIFF
--- a/basic/token2.s
+++ b/basic/token2.s
@@ -172,7 +172,7 @@ errtab	.word err01
 	.word err30
 
 okmsg	.byt $d,"OK",$d,$0
-err	.byt $20,"ERROR",0
+err	.byt " ERROR",0
 intxt	.byt " IN ",0
 reddy	.byt $d,"READY.",$d,0
 erbrk	=30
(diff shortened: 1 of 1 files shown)


basic/token2.s
```diff
@@ -172,7 +172,7 @@ errtab	.word err01
 	.word err30
 
 okmsg	.byt $d,"OK",$d,$0
-err	.byt $20," ERROR",0 ;add a space for vic-40 screen
+err	.byt $20,"ERROR",0
```
greg-king5, line 175: The `$20,` is the extra space (it's a `$0D` in VIC-20 BASIC).  You should remove it instead of the space in `" ERROR"`.

PR #208, opened 2021-04-17
Feature: Possible BASIC command to start a program stored in ROM

stefan-b-jakobsson:
In case the Commander X16 will allow users to store their own programs in ROM, it would be nice to have a BASIC command to start such programs. 

Without such a command, the user needs to load startup code from the SD card or enter a small BASIC or machine language program every time the computer is started. That would probably deter most users from ever wanting to run custom ROM based programs.

The code works, but is only a proof of concept, and not ready to be merged in my opinion.

A short summary of how it works:

- The code adds a new BASIC command START. You enter for instance START "X16EDIT" to start the editor I've been working on.
- The BASIC command reads the string specified in the command, and calls a new KERNAL function ROM_START
- ROM_START searches all ROM banks for the program. If found, it is started.

The function works only with ROM banks formatted in a standardized way:
- $C000-C002 must contain a JMP instruction to the programs main entry point
- $C003-C005 must contain the magic string "X16"
- $C006-C008 may contain program version, major-minor-patch
- $C009 must contain the length of the program name
- $C00A and forward max 255 bytes must contain the program name

This format is a combination of what the community would like according to this thread:  https://www.commanderx16.com/forum/index.php?/topic/956-signature-for-rom-based-programs/

All ROM bank not meeting these conditions are ignored in the search.

Even though the code works, I suppose there are some considerations:

- I don't know if it will be practical or possible for users to add custom programs to the ROM. If not, this feature is of no practical use and should not be merged
- It consumes one KERNAL vector. Maybe it could be handled in another way
- It uses R0 and R1 zero page registers. That's neither ideal in the production code.
- I am no expert on the layout of the X16 KERNAL project, and it's likely that Michael would want to change it a lot

Even so, it's works nicely, and it takes away the need for constant pressure to include yet another special BASIC command just to start a ROM based program.


stefan-b-jakobsson:
Hi,

I have made a small ROM test image as follows (two files):

First the file start.s.

`banked_irq          = $038b ; irq handler: RAM part
VERA_L              = $9f20
VERA_M              = $9f21
VERA_H              = $9f22
VERA_D0             = $9f23

.CODE

;HEADER
jmp main_entry
.byt "x16"      ;Magic string
.byt 1, 0, 0    ;Program version
.byt 4          ;Program name length
.byt "test"     ;Program name

;CODE
main_entry:
    stz VERA_L
    lda #20
    sta VERA_M
    lda #(2<<4)
    sta VERA_H
    ldy #0

:   lda hello,y
    beq exit
    sta VERA_D0
    iny
    bra :-

exit:
    rts

hello:
    .byt "hello, world!", 0

.segment "IRQ"
    .byt $ff, $ff, $ff, $ff, <banked_irq, >banked_irq`


And then a config file test.cfg.

`MEMORY {
    ZP:             file = "", start = $0022, size = $0080 - $0022, define = yes;
    ROMBANK:        start = $c000, size = $3ff0, fill=yes, fillval=$aa;
    VECT:           start = $fffa, size = $06, fill=yes, fillval=$aa;
}
SEGMENTS {
    ZEROPAGE: load = ZP,        type = zp, define=yes;
    CODE:     load = ROMBANK,   type = ro;
    IRQ:      load = VECT,      type = ro;
}`


Compile with: cl65 -v -o test.bin -t cx16 -C test.cfg --mapfile test.map -vm  test.s

stefan-b-jakobsson:
I also made minimal ROM test image that works with the START command.

It consists of the following two files:

https://github.com/stefan-b-jakobsson/x16-rom/blob/ROM_START/test.cfg
https://github.com/stefan-b-jakobsson/x16-rom/blob/ROM_START/test.s

Compile with: cl65 -v -o test.bin -t cx16 -C test.cfg --mapfile test.map -vm  test.s

To append the resulting ROM image, type "cat rom.bin test.bin > custom.bin" (if using Linux/MacOS, where rom.bin is the standard ROM image.

PR #213, opened 2021-05-11
Feature: FSYS, a command to run code in another ROM bank

stefan-b-jakobsson:
The FSYS BASIC command may be used to run code stored in one of the ROM banks.

Even though I understand that the ROM is not primarily intended for user programs, a lot of users will probably anyway want to store different utilities there. 

Without a BASIC command to run ROM based code, a startup stub must be manually entered into RAM.

Usage: FSYS [ADDRESS],[ROM BANK]

Example: FSYS $C000,7 => runs code in ROM bank 7 at address $C000


stefan-b-jakobsson:
I'm closing this PR as it's not likely to be accepted.
A separate BANK statement could be a possible solution to the problem.

PR #216, opened 2021-08-07
Add support for headerless files

### LINKED ISSUE #211: Feature Request: Loading headerless files
The kernal LOAD routine expects files to contain a two-byte header specifying the address at which they should be loaded. Although the LOAD routine can be instructed to load at an arbitrary address, it is impossible to load the first two bytes of a file as data using only this routine. This makes it difficult to handle file formats that do not include this header.

Elektron72:
This pull request gives the kernal load routine the ability to load files that do not have a two-byte address header. This is done by interpreting the secondary address as described below:

<table>
	<tr>
		<th>Name</th>
		<th>Bit&nbsp;7</th>
		<th>Bit&nbsp;6</th>
		<th>Bit&nbsp;5 </th>
		<th>Bit&nbsp;4</th>
		<th>Bit&nbsp;3 </th>
		<th>Bit&nbsp;2</th>
		<th>Bit&nbsp;1 </th>
		<th>Bit&nbsp;0</th>
	</tr>
	<tr>
		<td align="center">SA</td>
		<td colspan="6" align="center">Unused</td>
		<td align="center">Write Header</td>
		<td align="center">Address Source</td>
	</tr>
</table>

If **Address Source** is 0, the load address is taken from the X/Y registers. If **Address Source** is 1, the load address is taken from the file header. Using the first bit in this way allows for partial backwards compatibility with previous versions of the load routine.

If **Write Header** is 0, the header bytes are not written into memory. However, if **Write Header** is 1, the header bytes are written into memory.

Closes #211.

### DIFF
--- a/kernal/cbm/channel/load.s
+++ b/kernal/cbm/channel/load.s
@@ -14,8 +14,13 @@
 ;* determined by contents of      *
 ;* variable fa.                   *
 ;*                                *
-;* alt load if sa=0, normal sa=1  *
-;* .x , .y load address if sa=0   *
+;* sa byte:                       *
+;*  76543210                      *
+;*        ||                      *
+;*        |+-address source       *
+;*        +--write header         *
+;* .x , .y load address if        *
+;*   address source = 0           *
 ;* .a=0 performs load to ram      *
 ;* .a=1 performs verify           *
 ;* .a>1 performs load to vram;    *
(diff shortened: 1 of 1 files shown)


mist64:
I love it!

irmen:
Yay!!

PR #223, opened 2021-10-02
[BASIC] save bank setting in VPOKE and VPEEK so it won't be lost;

### LINKED ISSUE #167: VPOKE, VPEEK does not work correctly R38
Hi if you put in the following code, it does not work as expected:
```
10120 VPOKE 1,$F000+I,VPEEK(0,$F800+I)
```
The above statement is a part of a loop to move the CHARROM to another addressspace in VRAM.

However it does not copy to $1F000 - instead it copies to $0F000. It looks like the VPOKE does not set the address correctly in this example and keeps the MSB at what is given in the VPEEK instruction.

This code for example works, but there should be really no difference between both versions.
``` 
10115 BT=VPEEK(0,$F800+I)
10120 VPOKE 1,$F000+I,BT
```

Example code to demonstrate (you can see the result on the screen):
```
10000 :REM SETUP SCREEN SCALING 2X = 320X240
10010 POKE $9F2B,64:POKE $9F2A,64
10020 COLOR 1,0:PRINT CHR$(147);
10030 REM SETUP BITMAP MODE 8BPP AND BITMAP START $4000
10040 POKE $9F2F,$20:POKE $9F2D,$07:POKE $9F29,PEEK($9F29) OR $10
10100 :REM MOVE CHARROM
10110 FOR I=0 TO $7FF
10120 VPOKE 1,$F000+I,VPEEK(0,$F800+I)
10130 NEXT
10140 POKE $9F36,$F8
```

**Be aware the last command in this example scrambles the CHARROM as it is not where it is expected and carries  random data. So you will not be able to see anything anymore. ...** 

Working code
```
10000 :REM SETUP SCREEN SCALING 2X = 320X240
10010 POKE $9F2B,64:POKE $9F2A,64
10020 COLOR 1,0:PRINT CHR$(147);
10030 REM SETUP BITMAP MODE 8BPP AND BITMAP START $4000
10040 POKE $9F2F,$20:POKE $9F2D,$07:POKE $9F29,PEEK($9F29) OR $10
10100 :REM MOVE CHARROM
10110 FOR I=0 TO $7FF
10115 BT=VPEEK(0,$F800+I)
10120 VPOKE 1,$F000+I,BT
10130 NEXT
10140 POKE $9F36,$F8
```

Jaxartes:
In VPEEK() and VPOKE, saving the bank selection on the stack, and not writing it to VERA_ADDR_H until *after* the address (and, for VPOKE, value) is computed.  Without that, VPEEK() within VPOKE or another VPEEK() between two banks won't work right.

The most likely use case (the one reported in #167) is `VPOKE B1, A1, F(VPEEK(B2, A2))` but `VPOKE B1, F(VPEEK(B2, A)), V` and `VPEEK(B1, F(VPEEK(B2, A)))` would also have issues.  (Writing `F(X)` as shorthand for "some expression involving `X`".)

This fixes #167 .

On the downside, it adds 1.9 microseconds to VPEEK() and VPOKE; but that's only 0.4%.

Attached: two test programs:
[issue-167-3.bas.txt](https://github.com/commanderx16/x16-rom/files/7270840/issue-167-3.bas.txt)
[issue-167-4.bas.txt](https://github.com/commanderx16/x16-rom/files/7270841/issue-167-4.bas.txt)

### DIFF
--- a/basic/x16additions.s
+++ b/basic/x16additions.s
@@ -186,8 +186,7 @@ byte_to_hex_ascii:
 vpeek	jsr chrget
 	jsr chkopn ; open paren
 	jsr getbyt ; byte: bank
-	txa
-	pha
+	phx
 	jsr chkcom
 	lda poker
 	pha
@@ -208,8 +207,7 @@ vpeek	jsr chrget
 
 ;***************
 vpoke	jsr getbyt ; bank
-	txa
-	pha
+	phx
 	jsr chkcom
 	jsr getnum
 	pla
--- a/basic/x16additions.s
+++ b/basic/x16additions.s
@@ -186,7 +186,8 @@ byte_to_hex_ascii:
 vpeek	jsr chrget
 	jsr chkopn ; open paren
 	jsr getbyt ; byte: bank
-	stx VERA_ADDR_H
+	txa
+	pha
 	jsr chkcom
 	lda poker
 	pha
@@ -199,15 +200,20 @@ vpeek	jsr chrget
 	sta poker + 1
 	pla
 	sta poker
+	pla
+	sta VERA_ADDR_H
 	jsr chkcls ; closing paren
 	ldy VERA_DATA0
 	jmp sngflt
 
 ;***************
 vpoke	jsr getbyt ; bank
-	stx VERA_ADDR_H
+	txa
+	pha
 	jsr chkcom
 	jsr getnum
+	pla
+	sta VERA_ADDR_H
 	lda poker
 	sta VERA_ADDR_L
 	lda poker+1


Jaxartes:
I've changed the TXA / PHA to PHX, thanks.

As for clearing the ADDR_SEL bit, I'm thinking it's a separate problem, to be solved by someone who has more of a vision about how the ROM code _should_ work than I have.  One way it could go is, as you say, zero before use; another would be, zero or restore after use.  Currently, code which writes VERA_CTRL in the ROM seems to do the latter, setting it to 0 after 1 or 2 (the 2 being another bit, but similar story).  Example FB_filter_pixels in framebuffer.s.

PR #225, opened 2021-12-05
Fixed LOAD into banked RAM

ZeroByteOrg:
This patch fixes the behavior of DOS to properly consider banked RAM when performing block copies.
It adds a bank-aware routine which copies from the buffer into the bank that was active when DOS was called. If a bank wrap occurs, the saved RAM bank is incremented so that it will be active when DOS exits, and the fat32_ptr is also adjusted to point to the new "next byte" of RAM. The return value of n-bytes-transferred is unaltered.

Kernal LOAD routine was also altered to properly do a bank wrap if the initial LOAD was started into a destination < $C000.
I made sure not to use any 65c02-specific opcodes in this module, as it appears to be remaining C64-compatible.

LOAD still properly returns the address of the first byte following the bank-wrapped load.

### DIFF
--- a/dos/fat32/fat32.s
+++ b/dos/fat32/fat32.s
@@ -14,6 +14,8 @@
 	.import filename_char_ucs2_to_internal, filename_char_internal_to_ucs2
 	.import filename_cp437_to_internal, filename_char_internal_to_cp437
 	.import match_name, match_type
+	
+	.importzp bank_save, krn_ptr1
 
 	; mkfs.s
 	.export load_mbr_sector, write_sector, clear_buffer, set_errno, unmount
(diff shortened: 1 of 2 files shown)


ZeroByteOrg:
This PR adds 136 bytes to the DOS bank. I opted to go for speed and maintain the "count down" loop behavior, which complicated the bank wrap logic considerably. A slower "count-forward" loop may save some space. I will try this and if successful and savings are significant, I will make a second PR with that fix and the maintainers can decide which solution is preferable.

ZeroByteOrg:
I am going to close this pull request....

ZeroByteOrg:
Please use PR #226 instead.

PR #226, opened 2021-12-06
Fixed LOAD into banked RAM - smaller patch

ZeroByteOrg:
This patch fixes the behavior of DOS to properly consider banked RAM when performing block copies.

It adds a bank-aware subroutine to fat32_read in fat32.s which copies from the buffer into the RAM bank that was active when DOS was called by LOAD. If a bank wrap occurs during the block copy, bank_save is incremented so that the bank containing the next available byte of memory will be active when DOS exits. The destination file pointer fat32_ptr is also adjusted to point to the adjusted "next byte" of RAM. The return value of n-bytes-transferred is unaltered.

The Kernal LOAD routine was also altered to properly do a bank wrap if macnt returns a byte count that causes the data pointer to rise past $C000. When patching this module, I made sure not to use any 65c02-specific opcodes, as it appears to be remaining C64-compatible. Lastly, I re-enabled the previously disabled support for bank wraps in byte-by-byte mode.

LOAD still properly returns the address of the first byte following the bank-wrapped load.

This Pull request should be considered as superseding PR#225, as it is over 60 bytes smaller than the previous patch, and is only about 3 frames slower to copy 96KB into banked RAM. I am going to cancel that PR, but in case it remains visible to the maintainer, I wanted to mention that here.

### DIFF
--- a/dos/fat32/fat32.s
+++ b/dos/fat32/fat32.s
@@ -14,6 +14,8 @@
 	.import filename_char_ucs2_to_internal, filename_char_internal_to_ucs2
 	.import filename_cp437_to_internal, filename_char_internal_to_cp437
 	.import match_name, match_type
+	
+	.importzp bank_save, krn_ptr1
 
 	; mkfs.s
 	.export load_mbr_sector, write_sector, clear_buffer, set_errno, unmount
@@ -22,6 +24,8 @@
 FLAG_IN_USE = 1<<0  ; Context in use
 FLAG_DIRTY  = 1<<1  ; Buffer is dirty
 FLAG_DIRENT = 1<<2  ; Directory entry needs to be updated on close
+
+ram_bank = 0		; RAM banking control register address
 
 .struct context
 flags           .byte    ; Flag bits
@@ -2957,6 +2961,10 @@ fat32_read_byte:
 ; * c=0: failure; sets errno
 ;-----------------------------------------------------------------------------
 fat32_read:
+
+	tmp_swapindex	= krn_ptr1   ; use meaningful aliases for this tmp space
+	tmp_done		= krn_ptr1+1 ; during bank-aware copy routine
+
 	stz fat32_errno
 
 	set16 fat32_ptr2, fat32_size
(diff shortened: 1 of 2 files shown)


ZeroByteOrg:
My explanation of the exit condition of a banked LOAD is a little wordy and unclear. Here's an example:
MYFILE = exactly 8K file size, and the active RAM bank is 1.

LOAD "MYFILE",8,3,$A000
LOADING FROM $A000 TO $A000
READY.
[]
The active RAM bank will be 4. The data will be in bank 3 as specified in the LOAD command.
This behavior is also correct for standard LOAD "FILE",8,1 if FILE's 2-byte load header is in the banking window.

ZeroByteOrg:
I should add this important note: My patch causes LOAD to clobber Kernal ZP register R15L ($20). This isn't strictly NECESSARY to do the job, I did this in order to cause LOAD to behave normally when writing to addresses >$c000 in order not to wrap such writes back down into HiRAM.

mist64:
Thanks for your patch and sorry it took me so long to look at it.

I added some comments. In general though, I really like it, especially how it reports back the end address. It's very consistent. :)

I took your code, created a branch and did some more work on it: https://github.com/commanderx16/x16-rom/commits/p226

* I fixed everything I complained about in your changes. ;-)
* I got rid of the use of $20

It'll need some more work – and testing.

dos/fat32/fat32.s
```diff
@@ -3055,6 +3074,67 @@ fat32_read:
...
+	bne @nowrap
+	inc bank_save
+	inx
+	lda #$9f
+	sta fat32_ptr+1
+@nowrap:
+	cpy	tmp_done
+	bne @loop
+
+;	add16 fat32_ptr, fat32_ptr, bytecnt
```
mist64, line 3120: ???

mist64:
Actually, I think the `LOADING FROM` message should read something like:

    LOADING FROM $03:$A000 TO $04:$B234

in case of loading into banked RAM, i.e. the addresses should be prefixed with the bank, so it's less confusing, *and* it actually carries all information (in the example, $3234 bytes were loaded).

mist64:
Since this is a feature, not a bug, I'll have to tend to other issues at the moment. My code is in a public branch, if you (or anyone) wants to continue working on this in the meantime.

mist64:
(Upgraded this to "bug", because it doesn't conform to the specification.)

mist64:
I merged my version as https://github.com/commanderx16/x16-rom/commit/b6a0f0c3daca1ca3dd3e71abf70ba55260381efa

The commit message contains you in the "Co-author" field, I hope that's okay.

irmen:
Great to see this land! :tada:

Issue #234, opened 2022-03-23
MONITOR wraps around if ROM bank is > 7


stefan-b-jakobsson:
It seems like the built-in MONITOR wraps around if the ROM bank specified with the O command is greater than 7.

If you enter "O08" and then display memory with "M C000" you get the contents of ROM bank 0, and so on.

I made a small test with ROM bank 8 populated: a small assembly program copying the contents of ROM bank 8 to RAM. It worked fine. So the problem must be within the MONITOR code.

The problem in the MONITOR code wasn't obvious to me when I looked at it briefly.

EDIT: It hit me that the monitor could be using KERNAL function FETVEC. And sure enough, there is an AND #7 in kernal/drivers/x16/memory.s line 214 that probably should be removed.

mist64:
Thanks!

PR #235, opened 2022-03-23
Fix: Changed Kernal FETVEC function to support 32 ROM banks

stefan-b-jakobsson:
This pull request fixes issue #234

The FETVEC function did support only ROM banks 0-7, but is here changed to support ROM banks 0-31. According to the PRG, the X16 Edit will have 512 kB of ROM => 512 kB / 16 kB = 32 banks


stefan-b-jakobsson:
The pull request was updated to completely remove the ROM bank limitation, in the same way as was done in pull request #206 

Fixes #234

Issue #267, opened 2022-04-10
CodeX is broken in r39


mist64:

mjallison42:
PR #269 submitted to fix this (and other) CODEX defects.

PR #269, opened 2022-04-10
CODEX Post R39 omnibus bug fixes

mjallison42:
Bug fixes for CODEX, recover from R39 regressions, etc. Primarily answer isse #267, "CODEX is broken"


codex/inc/x16_kernal.inc
```diff
@@ -108,17 +115,33 @@
 	FILE_LA       = $028f
 	FILE_SA       = $0290
 	FILE_FA       = $0291
+   cscrmd        = $0261
```
mist64, line 118: you can call `screen_mode` with .C=1 to get this.

In general, these KERNAL addresses will keep breaking, is there a way to do without them? What APIs do you need?

kernal/drivers/x16/screen.s
```diff
@@ -26,6 +26,9 @@
 ; for monitor
 .export pnt
 
+; for codex
+.export cscrmd
```
mist64, line 30: This is not actually used, `cscrmd` is hard-coded in CodeX.

mist64 (changes requested):

codex/inc/x16_kernal.inc
mjallison42: Good code review! This is an old hanger on, switched to proper JSRFAR call. (docs need to be updated). Also remove LA/SA/FA, I need to reconsider how to preserve state during BRK handling. Any way to get current LA/SA/FA?
PR has been updated to remove these references.

kernal/drivers/x16/screen.s
mjallison42: Removed, thanks for keeping me honest.

PR #271, opened 2022-04-13
Fix joystick routine messing with PS/2 keyboard

akumanatt:
Also tsb/trb optimizations

### DIFF
--- a/kernal/cbm/serial.s
+++ b/kernal/cbm/serial.s
@@ -39,9 +39,8 @@ count	.res 1           ;$A5 temp used by serial routine
 	.segment "SERIAL"
 
 serial_init:
-	lda d1prb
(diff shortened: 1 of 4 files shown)


mist64:
I tested this and:
* wasd keyboard behaves the same as before (ok at 2 MHz, no recation at 8 MHz)
* Fujitsu keyboard behaves the same as before (ok at 2 and 8 MHz)
* Serial bus now hangs on `DOS"$`

mist64:
Reverting the changes in serial.s makes the serial bus work again. It seems the faster code broke the timing?

jburks:
Oscilloscope trace after applying this change shows very clean signaling. Attached image shows both a press and release for the letter Q (US Keyboard).
image

PR #284, opened 2022-04-17
Add CODEX makefile rule to bundle plugins

mjallison42:
This PR adds a rule to the CodeX makefile to bundle plugins and example programs into an SDCard image "codex.img". This disk image will be used by CodeX user/developers. 

The PR also fixes an unreported defect when loading debug information from some programs.


codex/Makefile
```diff
@@ -143,7 +143,7 @@ clean:
 	rm -f $(CX_OBJS)
 	rm -f test/*.o test/*.l
 
-$(BUILD_DIR)/codex.bin: $(CX_SRCS) $(CX_INCS) $(CX_OBJS) $(KSUP_VECS) $(CFG_DIR)/codex-x16.cfg
+$(BUILD_DIR)/codex.bin: $(CX_SRCS) $(CX_INCS) $(CX_OBJS) $(KSUP_VECS) $(CFG_DIR)/codex-x16.cfg cx
```
mist64, line 146: This seems to reintroduce the old build problem?

mjallison42: New commit added, removes target "cx". I changed PR to draft (don't know if that was the correct thing to do). Should be good to go.

PR #289, opened 2022-04-27
P222 vmemcpy

### LINKED ISSUE #222: Add RAM/VRAM memcpy
GIVEN: There are two reasonably abstract sprite calls in the extended KERNAL.

1. Now that VERA has a PSG, it seems that EVENTUALLY the KERNAL could have a call or two to set up and play sounds.

2. It also seems reasonable (??) that EVENTUALLY the KERNAL could have a call to copy memory from system RAM into VRAM.  Or extend its current mem copy routine.

Because the KERNAL is an abstraction layer for "stuff", allowing implementations to vary without breaking code.

EXAMPLE 1.  The X8 has a different VERA access mechanism.  The KERNAL calls would disguise this to some degree.

EXAMPLE 2.  Some fool replaces VERA with VULKAN or something else.  He implements the KERNAL calls accordingly.  Code which uses the KERNAL might not care.

Example 2 is likely much trickier than I assume, but you get the idea.

mist64:
Documented here:
https://github.com/commanderx16/x16-docs/commit/de8cc7e56ab0768d26e774e7d00ebb02a7d416cd

Closes #222

### DIFF
--- a/kernal/memory.s
+++ b/kernal/memory.s
@@ -12,6 +12,8 @@
 
 .segment "MEMORY"
 
+IO_PAGE = $9f
+
 ;---------------------------------------------------------------
 ; memory_fill
 ;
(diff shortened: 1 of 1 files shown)


pglewis:
I tested this utilizing three different areas of VRAM: a custom palette, and two tile areas (one displayed, the other a draw buffer).  Everything worked as expected.

Issue #288, opened 2022-04-27
BASIC LOAD command syntax for headerless load


ZeroByteOrg:
Currently, LOAD command does not properly expose the new "headerless load" functionality.
Currently, the command takes two forms:
LOAD "FILE",device[,sa]
LOAD "FILE",device,bank,address

The first form behaves as: if (sa & 1) then "use PRG header as load-to address" else "load to $801"
If the user specifies an SA with bit 1 set, headerless mode is engaged. If sa & 3 is true, then the expected-but-strange behavior is to interpret the first two bytes as a PRG header and load the file to that address, but retain the first 2 bytes.
This should probably be an ILLEGAL QUANTITY ERROR instead?
(load with SA=2 seems useful if not often - so that should just work as-is)

I propose that the second form above should also get the optional [,sa] argument supported:
LOAD "FILE",device,bank,address[,sa] to give access to headerless load into specified address, which makes much more sense.

irmen:
I personally prefer a new BLOAD command reather than obfuscating the existing LOAD command more. Then we can leave out the 'sa' parameter altogether as well.

ZeroByteOrg:
either/or - I think BLOAD makes sense as it implies headerless and doesn't require the user to know about bits and masks and such. If implemented, I wonder how hard it would be to make legacy LOAD mask off the headerless bit to further enforce the feature, save users from themselves....

ZeroByteOrg:
I think a better name might be LOADBIN or BINLOAD as those names are a tad more descriptive than BLOAD.
Thinking of a way to hook the existing LOAD routine from such an "auxilliary call" such that the "headerless bit" is preserved in all cases...

ZeroByteOrg:
I decided to just go ahead and make the commands be B+[command] for consistency.
I have provided BASIC commands BLOAD, BVLOAD, and BVERIFY
See the comments on PR #312 for a couple of bugs in headerless load that I found while testing this PR.

PR #312, opened 2022-05-06
New BASIC commands BLOAD BVLOAD BVERIFY

ZeroByteOrg:
This addresses issue #288 
These BASIC tokens act the same as LOAD, VLOAD, VERIFY except that they require the extended syntax of "FILENAME",device,bank,address

Known issues:
While testing, I discovered that the Kernal's headerless load support has two bugs that affect the behavior of these BASIC tokens:
- headerless VLOAD does not do `eal += 2` when headerless mode is detected. Thus the reported final address is 2 bytes too low when complete.
- headerless mode verify does not verify the first two bytes, and does not increment eal by two. Thus BVERIFY will fail against the original file/address. If the user specifies file, with address+2, then bverify returns OK (presuming that the data is in fact correct)
- Patching the Kernal load routine will fix these issues with these tokens' behavior without any changes to them.

I attempted to patch the Kernal load routine, but there is not enough space left in the ROM bank to fit any fixes. This should probably be filed as a new issue (the buggy headerless load, not the full rom bank)


basic/code26.s
```diff
@@ -174,9 +175,14 @@ jerxit	jmp erexit
 ;- the carry flag is clear
 ;- eormsk contains 1
 ;
+plsvbin
+	lda #2			; enable headerless mode
+	.byte $2c		; skip lda #0 ($4c = bit xxxxx)
```
theelkmechanic, line 180: Picking nits, but the opcode in the comment is wrong. :-)

ZeroByteOrg: lol - who reads comments? I'll fixx0r and push another update to this PR. :)

basic/code26.s
```diff
@@ -174,9 +175,14 @@ jerxit	jmp erexit
...
 plsv
+	lda #0			; disable headerless mode
 ;default file name
 ;
+	sta addend		; headerless mode stashed in addend
```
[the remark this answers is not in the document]
ZeroByteOrg: fixt also

basic/x16additions.s
```diff
@@ -225,9 +225,26 @@ vpoke	jsr getbyt ; bank
 	rts
 
 ;***************
+bvrfy
+	lda #1
+	.byte $2c ; skip lda #0 (bit xxxx)
```
mist64, line 230: we can't use the $2C trick on the X16 because the BIT command could read from I/O registers, which may have side effects. We replaced all instances in KERNAL and BASIC with `BRA`.

basic/code26.s
```diff
@@ -174,7 +175,12 @@ jerxit	jmp erexit
 ;- the carry flag is clear
 ;- eormsk contains 1
 ;
+plsvbin
+	lda #2			; enable headerless mode
+	.byte $2c		; skip lda #0 ($2c = bit $xxxx)
```
mist64, line 180: instead of skipping the `lda #0`, you could have an entry point at line 183 and call it with .A=2

mist64:
I made the requested changes myself and fixed some spacing, and merged it as d383630. :)

Issue #313, opened 2022-05-06
Headerless load has bugs in vram / verify behaviors


ZeroByteOrg:
1) When performing a headerless load into VRAM, the ending address is reported incorrectly. (it is 2 bytes below the actual ending address)

2) When performing a verify operation and headerless mode is set, the routine does not verify the initial 2 bytes, and does not update the eal/eah pointer, so it falsely reports a failed verification. If you perform a verify with the [address+2] to skip the missing two bytes, then it returns the proper result.

I had made a fix for item 1, but there is only 8 bytes of space available for more code in the Kernal ROM, as my patch there was only 12 bytes and overflowed the bank.

I have opted to report this issue as a known issue so that it can be tracked while Kernal ROM space issues are resolved otherwise.

PR #314, opened 2022-05-07
Fix headerless load for verify/VRAM cases

theelkmechanic:
This PR addresses issue #313

The previous headerless load additions to the kernal load routine would always read the first two bytes of the file into memuss, and then later if headerless load was requested, they would pull those bytes out and pretend they were just read. Two issues found with this when adding commands to BASIC for headerless load support:

1. When performing a headerless load into VRAM, the ending address is reported incorrectly. (it is 2 bytes below the actual ending address)
2. When performing a verify operation and headerless mode is set, the routine does not verify the initial 2 bytes, and does not update the eal/eah pointer, so it falsely reports a failed verification. If you perform a verify with the [address+2] to skip the missing two bytes, then it returns the proper result.

This PR fixes both issues by stashing the headerless load flag in TMP2 (which appears to be unused during the load process), then if headerless load is requested, it does not read the first two bytes at all; rather it checks after the read to see if it is the first one, and if so, checks for the file not found error then. This not only fixes both of the above issues, it saves 13 bytes of code space in the process.


mist64:
Sorry to nitpick, but can you, for consistency with the existing code, please use spaces to indent the comments (but keep tabs for indenting the instructions)? :)

Issue #310, opened 2022-05-06
Enable `smc_set_activity_led`


mist64:
The function `smc_set_activity_led` is currently a no-op, it needs to be reenabled and tested.

mist64:
If the `rts` is removed, the whole I2C system seems to get out of sync, and `PRINT DA$` is broken.

mist64:
Maybe the first call to this is earlier than the DDR setup, which messes up the sync.

PR #324, opened 2022-05-22
Clear Z flag before leaving sda_low

### LINKED ISSUE #323: Cannot set RTC due to Z flag semantics in i2c_write
In #271 the `sda_low` function was changed to use a `tsb` instruction instead of an `ora`/`sta` pair. It is possible for `i2c_write` to branch to `rec_bit` and then fall through to `sda_low` which can ultimately cause `i2c_write` to return with the Z flag set using different semantics than expected.

This issue is causing `i2c_write_byte` to branch to its error exit point when no error has occurred. This is the root cause of failures to set the time on the real-time clock on r40 and r41.

jburks:
This PR fixes #323 by ensuring Z flag is clear when returning from `sda_low`.

Tested on hardware where RTC could not have its time set on r41; RTC time changes (e.g. `TI$="235959"` from BASIC) correctly take effect after this change.

### DIFF
--- a/kernal/drivers/x16/i2c.s
+++ b/kernal/drivers/x16/i2c.s
@@ -189,6 +189,7 @@ rec_bit:
 sda_low:
 	lda #SDA
 	tsb ddr
+	ina             ; clear Z
 	rts
 
 i2c_stop:


jburks:
This also seems to fix #310 on my hardware.

PR #331, opened 2022-07-14
Improve I2C support and move PS/2 keyboard support to SMC

jburks:
This change set provides improvements to the kernel I2C support. Changes include:

- I2C pins moved to VIA1 PA0/PA1 (previously used for PS/2 keyboard)
- Supports I2C clock stretching and high I2C wire capacitance / weak pull-ups by waiting for SCL to read logical 1 on every call to `scl_high`
- Cleans up the I2C signaling as viewed on a scope / analyzer.
- Minimizes almost all unnecessary delays; only i2c_start and i2c_stop still have delays, which may not be strictly necessary.
- Optional inlining of I2C functions for a 43% reduction in keyboard scan time.

The throughput at 8MHz is equivalent to an average I2C clock of approximately 150KHz. Fetching a keyboard scan code (two transactions of two bytes each) takes approximately 284us when using inlined functions and 500us when using JSR instructions.

Using community-written I2C testing tools, roughly 250K I2C scan code request transactions were performed over roughly two hours with zero detected errors. The software test tools could detect error returns from `i2c_write_byte`, `i2c_read_byte` and also mismatched return value when performing read backs from the SMC echo test register.

This PR removes SystemFont from the kernel BANK but does place it elsewhere. This was needed because the new I2C code causes the kernel to overflow bank 0 by about 41 bytes.


kernal/drivers/x16/i2c.s
```diff
@@ -7,8 +7,8 @@
```
mist64, line 7: please add names to the Copyright in line 4.

jburks:
Closing this until Stefan's PR that properly clears up bank 0 kernel space goes through.

PR #337, opened 2022-10-13
Fast VLOAD

ZeroByteOrg:
This speeds up VLOAD by enabling block reads directly into VRAM in the LOAD routine using MACPTR. This is over ten times faster than the previous byte-by-byte loading that was done.

This requires a potentially-breaking change for software that uses MACPTR directly. MACPTR now uses the C flag to select between normal loading behavior, and a new copy loop which does not increment the memory destination pointer as data is read from the buffer.

I have implemented the new behavior by creating a new copy loop in fat32_read. The new loop is required, since the previous copy loop uses count-down-to-zero behavior on the loop index, which would reverse the order of bytes being sent into VRAM. Modifying the existing loop would have also required a check on each iteration whether to increment the dst counter, which would be slower, so I opted for a purpose-built copy loop instead.

Kernal's own calls to MACPTR are patched to work with the new API functionality on MACPTR, but end-user code that calls MACPTR will need to ensure the carry flag is clear prior to calling MACPTR or else it will cause data loading to behave incorrectly.


ZeroByteOrg:
Accompanying emulator patch is PR#432

dos/fat32/fat32.s
```diff
@@ -3056,6 +3061,7 @@ fat32_read:
 @6c:
 	; fat32_ptr += bytecnt, fat32_bufptr += bytecnt, fat32_size -= bytecnt, file_offset += bytecnt
 	add16 fat32_ptr, fat32_ptr, bytecnt
+:
```
mist64, line 3064: name this @6c or something. This seems to be the target in line 3156?

dos/fat32/fat32.s
```diff
@@ -3068,7 +3074,8 @@ fat32_read:
 @7:
 	sec	; Indicate success
 
-@done:	; Calculate number of bytes read
+@done:
```
mist64, line 3077: undo?

dos/fat32/fat32.s
```diff
@@ -3114,23 +3121,45 @@ x16_banked_copy:
...
+	lda (fat32_bufptr),y
+	sta (fat32_ptr)
+	iny
+	dex
+	bne @loop
+@last:
+	lda (fat32_bufptr),y
+	sta (fat32_ptr)
+	jmp :-
+
```
mist64, line 3157: remove two blank lines

dos/fat32/fat32.s
```diff
@@ -3114,23 +3121,45 @@ x16_banked_copy:
...
+
 .endif
 
 
+
```
mist64, line 3162: undo

kernal/cbm/channel/load.s
```diff
@@ -106,13 +106,25 @@ ld30	jsr loding      ;tell user loading
 ;
 ;block-wise load into RAM
 ;
-bld10	jsr stop        ;stop key?
+bld10
```
mist64, line 109: undo?

dos/fat32/fat32.s
```diff
@@ -3035,6 +3035,11 @@ fat32_read:
 	; Copy bytecnt bytes from buffer
 	ldy bytecnt
 .ifdef MACHINE_X16
+.importzp krn_ptr1
+	bit krn_ptr1    ; test if C=1 or C=0 when called... 1=stream copy, 0=normal
```
mist64, line 3039: in fat32.s, it's not a question of C=1 or C=0 any more (that's at the MACPTR level), so this should just say: "MSB means..."

mist64 (commented):
just nits

mist64:
Once this is merged:
* https://github.com/commanderx16/x16-emulator/pull/432 will mirror this for the Host FS in the emulator
* we need to document it

dos/fat32/fat32.s
mist64: Ah I see, it's tricky to access a local variable from there. Still, "-" is suuuuper fragile!

ZeroByteOrg: I know- I tried various other workarounds and this is the only one that would work w/o having to refactor the labels entirely.

kernal/cbm/channel/load.s
ZeroByteOrg: no, I just moved the jsr stop and comment down to a new line to leave the label alone on this line

dos/fat32/fat32.s
ZeroByteOrg: done

dos/fat32/fat32.s
ZeroByteOrg: no, I just moved the comment down to be lined up with the lines of code that it is referring to.
I will move it back up...

dos/fat32/fat32.s
ZeroByteOrg: done

kernal/cbm/channel/load.s
ZeroByteOrg: "no" = "Was this an undo? No, I just...."
(clarifying what I meant by my response)

mist64:
I cleaned up the label magic in fce973e79f0930db770ef96cf74ff14061737f1e.

PR #341, opened 2022-10-16
Add WAI to text input busy loop

LRFLEW:
I was messing around with box16's added tools, and while looking at the CPU Visualizer, I realized that when you idle at the BASIC prompt, the system spends a lot of time in a busy loop. The emulator can handle the `WAI` instruction (commanderx16/x16-emulator#287), so I figured I'd try to update the ROM to use it. I'm not very familiar with the ROM project, so there might have been a better way to implement this, but I did test it and it did work for me. cc65 lacks a `WAI` mnemonic for the 65C02 (even though it has it for the 65816), so I had to use a `.byte` directive to add it.

### DIFF
--- a/kernal/cbm/editor.s
+++ b/kernal/cbm/editor.s
@@ -249,8 +249,10 @@ loop3
 .ifp02
 	beq loop3
 .else
+        ; power saving: a character from the keyboard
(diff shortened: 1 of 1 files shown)


mist64:
Thank you!

Issue #342, opened 2022-10-21
running PSET x,y,color while in text mode crashes BASIC into the monitor


mooinglemur:
Minimal case:

At the READY prompt:
`PSET 0,0,0`

which immediately enters the machine language monitor.

mooinglemur:
This also affects other drawing commands such as RECT and LINE

more appropriate outcomes could be

- no-op, silent failure
- enter graphics mode and apply the drawing command
- return an error such as `?ILLEGAL QUANTITY` or perhaps `?DEVICE NOT PRESENT`

mooinglemur:
happens in R40 and R41, unsure about current master particularly after the graph bank move.

Jaxartes:
I can confirm that this happens with the latest ROM code, which includes commit dd6623727b1fa846. It also happens with other graphics commands, such as "RECT0,0,100,100,100" or "LINE0,0,100,100,100".

The cause of the problem is that the graphics driver isn't initialized until you switch to graphics mode. If you switch to graphics mode then back to text mode then these commands won't crash, they just don't do anything visible. 

Some thoughts on coding a fix in ROM, if this "no-op, silent failure" behavior is desired:

1. Tried, doesn't work great: I tried modifying 'screen_mode' so as to initialize the graphics driver (through 'GRAPH_init') when entering text mode, too, and not just graphics mode. It worked, but also forced text mode to 40x30 which is bad.
2. Works, simple, feels a little lame: Modify the kernal's screen initialization ('cint') where it sets mode 0, to set mode 128 first. In my testing this seems to solve the problem. It might produce a brief video glitch on startup and reset but that may be acceptable (and consistent with some historic 8 bit systems).
3. Haven't tried: Create a "dummy" graphics driver which implements the same API as frambebuffer.s but doesn't do anything.

mist64:
> * no-op, silent failure

Not very user-friendly.

> * enter graphics mode and apply the drawing command

There are several modes, which one?

> * return an error such as `?ILLEGAL QUANTITY` or perhaps `?DEVICE NOT PRESENT`

I like this one. But we probably want a new error. The C128 prints `?NO GRAPHICS AREA ERROR`; I think we need better wording. `?NO GRAPHICS MODE` maybe?

Jaxartes:
New pull request https://github.com/commanderx16/x16-rom/pull/351 replacing the old one, to fix this by making an error.

PR #343, opened 2022-10-23
[KERNAL] Initialize graphics driver on startup (#342)

Jaxartes:
Does it by going to graphics mode before text mode.

That's kind of a dumb way to fix it, but it works in my testing. It has the potential to cause a brief video "glitch" on startup or reset. On startup I haven't noticed one (beyond what's already there). On reset it briefly goes to 40x30 text mode before going back to 80x60.

This assumes "silent failure" when issuing graphics commands in text mode is the right way to address issue #342. That's what currently happens after "SCREEN128:SCREEN0".

Test BASIC program attached, because I like to write these. 😀
[text-and-gfx.txt](https://github.com/commanderx16/x16-rom/files/9845678/text-and-gfx.txt)

### DIFF
--- a/kernal/cbm/editor.s
+++ b/kernal/cbm/editor.s
@@ -173,7 +173,12 @@ cint	jsr iokeys
 	jsr panic       ;set up vic
 
 	; XXX this is too specific
-	lda #0          ;80x60
+	; Screen mode: First 320x240 graphics mode (just to initialize
+	; the framebuffer driver) then 80x60 text mode.
+	lda #128        ;320x240 graphics
+	clc
+	jsr screen_mode
+	lda #0          ;80x60 text
 	clc
 	jsr screen_mode ;set screen mode to default
 ;


mist64:
Not a fan. Please see my comment in the issue.

Jaxartes:
Ok. I've been experimenting with adding such an error.

Calling "screen_mode" to query the mode works, but at a performance cost of ~40 microseconds -- which is about 6% for PSET. That doesn't seem good to me. What do you think?
(Attaching performance test program.)
[gfx-speed.bas.txt](https://github.com/commanderx16/x16-rom/files/9857374/gfx-speed.bas.txt)

Alternatives:

1. Look at "cscrmd" to get the mode. But that isn't BASIC's variable...
2. Store the mode set with "SCREEN" in a byte, and check that. But then if anyone uses some custom machine language to switch the screen mode it won't behave well.

mist64:
> Calling "screen_mode" to query the mode works, but at a performance cost of ~40 microseconds -- which is about 6% for PSET. That doesn't seem good to me. What do you think?

This doesn't sound too bad to me.

Jaxartes:
> > Calling "screen_mode" to query the mode works, but at a performance cost of ~40 microseconds -- which is about 6% for PSET. That doesn't seem good to me. What do you think?
> 
> This doesn't sound too bad to me.

That's good, because my attempts to code a faster alternative have all failed.

Working on testing the 'jsr screen_mode' code. Might have it ready 2 days from now.

Jaxartes:
Created a new pull request, with error instead of silent failure as discussed.
https://github.com/commanderx16/x16-rom/pull/351

PR #351, opened 2022-10-28
[BASIC] fail graphics commands in non graphics mode (#342)

Jaxartes:
with a new error message for it, "?NOT GRAPHICS MODE". This changes the commands PSET, LINE, RECT, FRAME, and CHAR.

The old behavior was:
PSET, LINE, RECT, and FRAME would crash to monitor, if graphics mode had never been entered. (CHAR didn't.) If graphics mode had been entered and exited, they appeared to do nothing.

This fix does impose a performance penalty, calling "screen_mode" with each operation. The attached test program [gfx-speed-2.bas.txt](https://github.com/commanderx16/x16-rom/files/9885186/gfx-speed-2.bas.txt)
showed a performance penalty of 40.5 microseconds per command -- 6.2% of the time for PSET (the fastest of the commands).

Some alternative ways of finding out the screen mode didn't turn out well.

### DIFF
--- a/basic/token2.s
+++ b/basic/token2.s
@@ -185,7 +185,7 @@ reddy	.byt $d,"READY.",$d,0
 erbrk	=30
 brktxt	.byt $d
 err30	.byt "BREAK",0,$a0 ;shifted space
-err31	.byt "NOT GRAPHICS MOD",$c5
+err31	.byt "NOT GRAPHICS MOD", 'E'+$80
 errngm	=31
 
 forsiz	=$12
--- a/basic/token2.s
+++ b/basic/token2.s
@@ -176,6 +176,7 @@ errtab	.word err01
 	.word err28
 	.word err29
 	.word err30
+	.word err31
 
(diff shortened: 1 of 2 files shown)


basic/token2.s
```diff
@@ -184,6 +185,8 @@ reddy	.byt $d,"READY.",$d,0
 erbrk	=30
 brktxt	.byt $d
 err30	.byt "BREAK",0,$a0 ;shifted space
+err31	.byt "NOT GRAPHICS MOD",$c5
```
mist64, line 188: The existing error strings couldn't do this because of limitations of the 80s assembler, but now we can write:

`.byt "NOT GRAPHICS MOD",'E'+$80`

mist64 (changes requested):
mini-nit.

basic/token2.s
mist64: (please only change this instance, not the existing ones to keep the patch small)

mist64:
Thanks!

Issue #352, opened 2022-10-28
BASIC overflow/hang in SQR()


Jaxartes:
In current code as of 27 Oct 2022 (latest commit f6a5bf599d2d0cd4) the following commands fail:
`A=2:B=1:C=0:X=-0.04062:Y=-0.04616:?Y-SGN(X)*SQR(ABS(B*X-C))`
`A=2:B=1:C=0:X=-0.04062:Y=-0.04616:?Y-SQR(ABS(B*X-C))`
The former gets "?OVERFLOW ERROR" while the latter just seems to hang (at least one minute without READY prompt).

Not clear how old the bug is (I think I need an older emulator to try older ROMs). It's not as old as Commodore 64 anyways.

Seen while computing a "Hopalong" fractal. Program attached.
[show-gfx.bas.txt](https://github.com/commanderx16/x16-rom/files/9885391/show-gfx.bas.txt)

Jaxartes:
The problem appears to start happening with commit ae608673f0210953.

Jaxartes:
The issue appears to be that when math functions were renamed from faddt to bfaddt, etc, not all calls to them were changed accordingly. In particular math/fsqr.s had calls to fdivt and faddt. And they do their own setup which is different from what "prepare" does.

I haven't fully tested this fix, which is why there isn't a pull request. But here it is in diff form:

1. The fsqr.s part only: 
[fix-352-diff.txt](https://github.com/commanderx16/x16-rom/files/9894636/fix-352-diff.txt)
2. And some other similar changes: 
[fix-352-diff-2.txt](https://github.com/commanderx16/x16-rom/files/9894637/fix-352-diff-2.txt)

I've tested both with the show-gfx.bas.txt program attached above, and both of the BASIC lines quoted above, and the problem seems fixed. But that's all I've done for testing so far on them.

MJoergen:
Have you tried the floating point test programs in https://github.com/commanderx16/x16-rom/tree/master/test/fp ? They may not be up-to-date, but they worked back-in-the-day. The idea is to run the same program with the old ROM and the new ROM, and compare the outputs.

Jaxartes:
Thanks @MJoergen for pointing out those tests. I've run them now and they give the same results with and without the fix (fix-352-diff-2.txt), and at commit 590a2bd10b18ac32. (Except for 1-2 jiffy differences in the "perf.bas" results.)

PR #355, opened 2022-10-30
[MATH] adjust calls and comments to functions renamed "bfaddt" etc

Jaxartes:
fixes issue #352 in which some uses of SQR() in BASIC hung or overflowed

Passes the "fp/" tests -- giving the same result as before. On the two one-liners given in #352, it doesn't overflow/hang but gives the same results as Commodore 64.


mist64:
Thank you!

PR #357, opened 2022-11-01
Update README.md

Yazwh0:
Python 3.7 (or above) is now needed to build the keyboard maps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Each [release of the X16 emulator][emu-releases] includes a compatible build of 
 
 ### Building the ROM
 
-Building this source code requires only [GNU Make] and the [cc65] assembler. GNU Make is almost invariably available as a system package with any Linux distribution; cc65 less often so. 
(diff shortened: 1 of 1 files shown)


mist64:
iirc Python isn't *required* for building. The keymap files are only generated on demand, the standard build only picks up the generated files. We should clarify that Python is only needed if one wants to re-generate the keymap files.

Yazwh0:
You can't just type `make` without it though, I guess the generated files aren't in the repo? (Its how I found that 3.6 wasn't new enough!)

mist64:
Oh let me verify…

mist64:
Hm, I'm confused. https://github.com/commanderx16/x16-rom/tree/master/keymap/asm contains the generated .s files. Can you remove Python 3.7 from your system again and show me the error you're getting on a clean build?

Yazwh0:
Don't need to uninstall Python to see it being used..

Unless of course something is checking if python exists before calling `./scripts/relist.py` and `./scripts/trace_lst.py`?

image

Yazwh0:
With no python installed:
image

mist64:
Aha! relist.pay needs it, not the keymap generators!

PR #358, opened 2022-11-01
modify CLRCHN to return prev. input file in .A and previous output file in .X

cnelson20:
- CLRCHN previously clobbered .A and .X so this change does not change compatibility.
- Currently there is no way to view the input / output file no's without peeking at kernal variables

### DIFF
--- a/kernal/open-roms/iostack/f333.clrchn.s
+++ b/kernal/open-roms/iostack/f333.clrchn.s
@@ -19,13 +19,16 @@ CLRCHN:
 	; FALLTROUGH
 
 clrchn_reset: ; entry needed by setup_vicii
-
-	; Set input device number to keyboard
-	lda #$00
-	sta DFLTN
-
+	
+	; Load previous output device num to .X
+	ldx DFLTO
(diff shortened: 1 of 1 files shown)


mist64:
Careful: There are two implementations of the core KERNAL functionality in this repo: Commodore's original code, and the "Open-ROMs" free reimplementation. Currently, we are using Commodore's code for the X16. The other version can be built, but I don't think it's currently functional. It is meant as a fallback for cases where the Commodore code can't be used for legal reasons.

The file you are changing is in the Open-ROMs code.

cnelson20:
> Careful: There are two implementations of the core KERNAL functionality in this repo: Commodore's original code, and the "Open-ROMs" free reimplementation. Currently, we are using Commodore's code for the X16. The other version can be built, but I don't think it's currently functional. It is meant as a fallback for cases where the Commodore code can't be used for legal reasons.
> 
> The file you are changing is in the Open-ROMs code.

Should I revert this change when I update my PR to edit the correct version of clrch?

PR #360, opened 2022-11-04
Build ROM in CI

maxgerhardt:
Get Github Actions to spin up a Ubuntu machine, clone, compile and install the latest cc65 toolchain from the git on it, make the ROM image, upload as an artefact. Very simplistic. See [here](https://github.com/maxgerhardt/x16-rom/actions).

The build time of ~1.5 minutes can be shortened if cc65 is installed from the package repos instead of bleeding edge, but I wanted to be sure the latest is used.

Correctly boots to the BASIC prompt when paired with the lates x16-emulator build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,27 @@
+name: ROM Build CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+      - name: Install Dependencnies
+        run: |
+          sudo apt-get install -y make build-essential
+          git clone https://github.com/cc65/cc65.git
+          cd cc65
+          make -j4
+          sudo make install
+      - name: Build ROM
+        run: make
+      - name: Archive firmwares results
+        uses: actions/upload-artifact@v3
+        with:
+          name: ROM Image
+          path: build/x16/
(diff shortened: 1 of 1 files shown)


indigodarkwolf:
Very cool stuff, I definitely think this is worth bringing in. Pinging @mist64 just to make sure he sees this and can provide feedback if he has time, otherwise I'll likely pull it in by the end of the weekend. :3

indigodarkwolf:
In fact, I wonder if Michael would like to see the build artifacts pruned a bit. This looks like it's capturing just about the entirety of the build process - every object file, every map, every piecewise bin file... might want to only capture `*.h` and `*.sym`, and `rom.bin`, and the main reason I suggest `*.sym` is because Box16 can look for and read those when they're next to the rom file.

maxgerhardt:
Sure, if intermediate files are not needed (thought it would be good for traceability / debug), we can copy the `*.sym`, `*.h` and `rom.bin` files on the top level (none of the netsted levels) in a new folder and use that as the artifact path.

maxgerhardt:
Also if instead of compiling the bleeding edge cc65 toolchain from the github we could use a fixed commit for the toolchain that every x16-rom developer may already be using. Let me know which one if it exists.

indigodarkwolf:
Oh, wait, seems I don't have merge permissions after all. 0:) Whoops. Umm... @mist64, can we merge this? Pretty please?

mobluse:
> Also if instead of compiling the bleeding edge cc65 toolchain from the github we could use a fixed commit for the toolchain that every x16-rom developer may already be using. Let me know which one if it exists.

On Raspbian GNU/Linux 10 (buster) I have from the built-in repo using apt:
cl65 V2.17 - Raspbian 2.17-1

PR #362, opened 2022-11-06
implement FB_set_palette

irmen:
Implements the FB_set_palette kernal routine.

A few remarks/questions:

- The index and count are interpreted to mean the *number of colors*, not a byte index. This allows access to all 256 colors and 512 byte palette locations, using only a single register for each. The documentation wasn't clear on this.

- is it okay to leave vera's autoincrement enabled at the end of the routine?

### DIFF
--- a/graphics/drivers/framebuffer.s
+++ b/graphics/drivers/framebuffer.s
@@ -73,22 +73,32 @@ FB_get_info:
 ;
 ; Function:  Set (a part of) the VERA's color palette
 ; Pass  :    r0       pointer to color palette data
-;            a        start byte index
-;            x        byte count
+;            a        VERA palette start color index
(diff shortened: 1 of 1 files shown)


mooinglemur:
I think the offset should be into the VRAM palette data, not an index into the source ptr r0.  Perhaps .A should be a color index rather than a byte index so you can reach all 512 palette data bytes with one index register?

I honestly don't know whether it's okay to use ADDR0/DATA0 and then leave it undefined for the next thing to use it to sort out.  My hunch is yes.

irmen:
I changed the implementation according to mooinglemur's comment.

Pass  :    r0       pointer to color palette data
           a        VERA palette start color index
           x        number of colors to set (0=256)

PR #364, opened 2022-11-08
reset VERA and YM in IOINIT kernal routine to silence sound on software reset

### LINKED ISSUE #330: reset doesn't silence the vera PSG
When vera PSG is making sounds, a RESET command issued on Basic, or a call to the system reset vector, does't silence the PSG.   Also writing 1 to the Vera RESET bit in $9f25 doesn't silence the PSG!  

Is this a kernal issue or a Vera issue? I assumed all vera registers would be reset to default state when resetting the vera?

pressing CTRL-R in the emulator *does* however, i don't really know why?

irmen:
This silences PSG sounds on system reset (or actually, re-initialization / software reset) and also puts the YM-2151 back into a clear state.

Fixes #330

### DIFF
--- a/kernal/drivers/x16/x16.s
+++ b/kernal/drivers/x16/x16.s
@@ -26,13 +26,60 @@
 ;---------------------------------------------------------------
 ioinit:
 	lda #%10000000
-	sta VERA_CTRL        ; reset Vera
+	sta VERA_CTRL        	; reset VERA
+	jsr ym_init		; reset YM
 	jsr vera_wait_ready
 	jsr serial_init
 	jsr entropy_init
 	jsr clklo       ;release the clock line
 	; fallthrough
 
+;---------------------------------------------------------------
+; Re-initialize the YM-2151 to default state (everything off).
+;
+;---------------------------------------------------------------
+YM_ADDRESS = $9f40
+YM_DATA = $9f41
+
+ym_init:
+	ldx #$e0
+	ldy #$0f
+@1:	jsr @ym_write	; disable all lfos $e0..$ff
+	inx
+	bne @1
+	ldx #$08
+	ldy #7
+@2:	jsr @ym_write	; set key off for all voices
+	dey
+	bpl @2
+	ldx #$01
+	ldy #$02
+	jsr @ym_write	; reset lfo
+	ldx #$19
+	ldy #$80
+	jsr @ym_write	; clear pmd
+	ldx #$0f
+	ldy #0
+@3:	jsr @ym_write	; clear everything else $0f..$ff
+	inx
+	bne @3
+	ldx #$01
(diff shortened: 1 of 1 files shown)


kernal/drivers/x16/x16.s
```diff
@@ -25,12 +25,61 @@
...
+	bpl @2
+	ldx #$01
+	ldy #$02
+	jsr @ym_write	; reset lfo
+	ldx #$19
+	ldy #$80
+	jsr @ym_write	; clear pmd
+	ldx #$0f
+	ldy #0
+@3:	jsr @ym_write	; clear everything else $0f..$ff
```
irmen, line 63: is this correct?  we already wrote $e0-$ff before

irmen:
In the Kernel channel on the discord, a hardware change was mentioned that would remove the need for the complicated YM reset routine.  It would add a feature that allows writing $40 into register $1B to causes a HW reset of the YM chip. But the PR here is for current hardware.

ZeroByteOrg:
I looked the code over and it looks good.
Question for Mist: would it make sense to expose the ym_init routine in the Kernal API? This would be useful for programs to be able to call in lieu of including similar code in memory that does the same thing. I can see a case either way - just wanted to throw the suggestion in the ring for consideration.

kernal/drivers/x16/x16.s
```diff
@@ -25,12 +25,72 @@
 ;            -- This is KERNAL API --
 ;---------------------------------------------------------------
 ioinit:
+	lda #%10000000
+	sta VERA_CTRL        	; reset VERA
```
irmen, line 29: Is this ok on real hardware? How does this interact with it when it is using composite display output?  (I think this will default back to VGA output)

I am *guessing* this doesn't really matter yet here because the kernal doesn't yet know about what output mode it should use anyway at this stage

ZeroByteOrg:
I'm also working on a proposed audio utility bank for the ROM, in which case I'll include YM and PSG init routines there and make the Kernal call that code instead, depending on whether this PR is merged first or not.

kernal/drivers/x16/x16.s
ZeroByteOrg: Dave wants to have the default screen output mode stored in NVRAM and set by the Kernal, which hasn't been implemented yet.

irmen: Right. So until that mechanism appears, I think it is okay to just reset to VGA because the system is doing that right now as well, I think?   (I can only see the behavior on the emulator but I guess the real hardware doesn't retain composite setting either when you reset it)

irmen:
Abandoning this PR in anticipation of a different PR with new audio routines that supersedes all the code here.

Issue #367, opened 2022-11-25
To match official Perixx keyboard keycaps, treat scroll lock as 40/80 switch in BASIC instead of (or in addition to) F4


mooinglemur:
The official Perixx keyboard (this image taken by The 8-Bit Guy himself) has a "40/80 DISPLAY" key in place of Scroll Lock.  I think we need to handle this key the same as the current behavior as F4, at least in the BASIC editor.

![x16perixx](https://user-images.githubusercontent.com/395186/204040663-3dd23455-72ed-406f-942e-3cc8136e2383.png)

Here is the same area of the generic Perixx keyboard for reference.

![normalperixx](https://user-images.githubusercontent.com/395186/204040679-7866d7a3-e708-473e-907d-5056003961d6.jpg)

mooinglemur:
Closed by https://github.com/X16Community/x16-rom/pull/3

PR #369, opened 2022-12-01
Caps Lock and 40/80 key support

stefan-b-jakobsson:
Summary of proposed changes:

- The code in this PR lights up the Caps Lock LED on the keyboard when Caps Lock is active.
- Support for multi-byte I2C write transactions. Currently it is only possible to write one byte at a time over I2C. Most keyboard commands require two bytes. The command to set keyboard LEDs on/off is an example of that. Even though it's possible to send commands to the keyboard over I2C one byte at a time, and waiting for keyboard ACK (0xFA) in between bytes, it's a lot more convenient to send commands in one single I2C transaction. This way the I2C_write functions resemble the I2C_read functions already merged into the Kernal.
- The Caps Lock behavior is changed. If you hold down the Shift key when Caps Lock is active you will get lower case characters. In other words, Caps Lock+Shift is unshifted default state
- Support for the 40/80 key (=Scroll Lock). The Scroll Lock key is ignored by the Kernal when not in the BASIC editor. The proposed code treats the 40/80 bit in shflag as a request to change screen mode. To prevent key presses during program execution to take effect immediately when returning to the BASIC editor, the 40/80 bit is always cleared at the label "main" in basic/code1.s.
- When the 40/80 flag in shflag is set the keyboard currently doesn't work. This has been fixed in the PR.
- The F4 key currently switches between 80 and 40 column modes. It's possible to use the key for another purpose. No change has been made in this PR.


mooinglemur:
This PR would address https://github.com/commanderx16/x16-rom/issues/367

stefan-b-jakobsson:
I've had some problems with either my SMC or this Kernal code. I'm marking it as draft again.

stefan-b-jakobsson:
I changed the status of this PR back to open/ready for review. According to my own tests, this Kernal code works properly. The main problems were in the SMC code, which also should be fixed.

basic/code1.s
```diff
@@ -35,7 +35,8 @@ readyx	lda #<reddy
 	lda #$80        ;direct messages on
 	jsr setmsg      ;from kernal
 
-main	jmp (imain)
+main	jsr clear_4080_flag
```
stefan-b-jakobsson, line 38: This clears the 40/80 bit in shflags on return to BASIC editor. It is intended to prevent 40/80 key presses during program execution to take effect on return to the editor.

kernal/cbm/editor.s
```diff
@@ -243,6 +249,35 @@ stdone
...
+	ldx #0          ;Print READY.
+:	lda redy,x
+	beq loop3b
+	jsr prt
+	inx
+	bra :-
+redy:	.byte "READY.", 13, 0
+
+loop3b:	pla
+	sta ram_bank
```
stefan-b-jakobsson, line 280: This part of the BASIC editor checks the 40/80 bit in shflag and toggles the screen mode between 80x60 and 40x30. The 40/80 bit is treated as a request to change screen mode.

kernal/cbm/editor.s
```diff
@@ -311,8 +346,9 @@ loop3a	jmp loop3
 ;
 lp29	pla
 	cmp #$d
-	bne loop4
-	ldy lnmx
+	beq :+
+	jmp loop4
+:	ldy lnmx
```
stefan-b-jakobsson, line 351: This change was necessary due to branch range.

kernal/drivers/x16/ps2kbd.s
```diff
@@ -264,15 +269,17 @@ _kbd_scan:
...
-	tay
 
+:	tay
 	cpx #0
 	jne down_ext
 ; *** regular scancodes
 	cpy #$01 ; f9
 	beq cycle_layout
-	cmp #$83 ; convert weird f7 scancode
+	cmp #$84 ; Eat weird Alt + PrtScr scan code
```
stefan-b-jakobsson, line 279: Ctrl+Alt+PrtScr (Restore) initiates an NMI. This is done in code running on the ATTiny. 
Normally PrtScr has the following PS/2 scancode (make/break): E0 12 E0 7C / E0 F0 7C E0 F0 12.
However, if Alt is pressed it has this scancode: 84 / F0 84
The Kernal currently seems to drop bit 7 of this, and interprets Alt+PrtScr as F3. To prevent this, the code above eats the PrtScr scancode.

kernal/drivers/x16/ps2kbd.s
```diff
@@ -283,14 +290,16 @@ _kbd_scan:
 	asl ; bit 6
 	php
 	lda shflag
-	and #<(~MODIFIER_CAPS)
+	and #(255-MODIFIER_4080-MODIFIER_CAPS)
```
stefan-b-jakobsson, line 293: The current code makes the keyboard not work if the 40/80 bit is set. Fixed by this change

kernal/drivers/x16/ps2kbd.s
```diff
@@ -283,14 +290,16 @@ _kbd_scan:
...
 	ror
 
-	ldx shflag
-	cpx #MODIFIER_CAPS
-	jeq handle_caps
+	tax
+	lda shflag
+	and #MODIFIER_CAPS
+	jne handle_caps
+	txa
```
stefan-b-jakobsson, line 302: This code calls handle_caps even if more bits than MODIFIER_CAPS are set.

kernal/drivers/x16/ps2kbd.s
```diff
@@ -554,6 +571,8 @@ receive_down_scancode_no_modifiers:
 	beq no_key
 	jsr joystick_from_ps2
 	php
+	cmp #$fa
+	bcs cmd_resp
```
stefan-b-jakobsson, line 575: Ignore PS/2 keyboard command ACK

PR #371, opened 2022-12-04
memory_decompress can decompress to VRAM

bsb-RickD:
decompression to VRAM did not work, because during decompression reads from already decompressed bytes (backreferences) are also needed.

To make it work I had to do the following changes:

1) the code implicitly assumes that you're decompressing to VERA ram and that the dataport 0 is the destination address.

2) had to make  add_match offset different for RAM and IO. The IO version of it now reads the current address from vera ADDR_L, M and H (on data port 0) does a 17 bit add and stores the calculated address again to vera, but on data port 1

3) had to make copy_backreference different for RAM and IO. The io variant reads from data1 and writes to data0.

With all these changes I then had 4 "function pointers" and needed to push 3 additional 16 bit registers onto the stack, and I didn't want that, so I changed the code to not use function pointers but to do the dispatch in each function based on the io bank - this is 2 cycles slower than the original code, but it results in smaller code, and since I now also use it for the putdst method  I actually now need to push 1 word less onto the stack.

### DIFF
--- a/kernal/lzsa.s
+++ b/kernal/lzsa.s
@@ -49,6 +49,16 @@ memory_decompress:
 memory_decompress_internal:
 	PushW r2
 
+	lda r1H
+	cmp #IO_PAGE
+	bne start_decompression
+	PushB VERA_CTRL                 ; we're decompressing to I/O - save VERA state
+	lda #1
+	sta VERA_CTRL                   ; switch to data port 1
+	PushB VERA_ADDR_L               ; save data port 1 address
+	PushB VERA_ADDR_M
+	PushB VERA_ADDR_H
+start_decompression:
 	ldy #$00
 	sty nibcount
 
@@ -213,7 +223,16 @@ combinedbitz:
 	rts
 
 decompression_done:
-	stz VERA_CTRL			; could also push/pull, but this is shorter, KERNAL assumes D0..
+	lda r1H
+	cmp #IO_PAGE			
+	bne dont_restore_io
+	lda #1                          ; we've decompressed to I/O - restore VERA state
+	sta VERA_CTRL                   ; switch to data port 1
+	PopB VERA_ADDR_H                ; pop the address for D1
+	PopB VERA_ADDR_M
+	PopB VERA_ADDR_L
+	PopB VERA_CTRL                  ; restore control
+dont_restore_io:
 	PopW r2
 	rts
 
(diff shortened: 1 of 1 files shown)


irmen:
> the code implicitly assumes that you're decompressing to VERA ram and that the dataport 0 is the destination address.

That sounds like a bad idea, now we cannot decompress to memory anymore?

bsb-RickD:
> > the code implicitly assumes that you're decompressing to VERA ram and that the dataport 0 is the destination address.
> 
> That sounds like a bad idea, now we cannot decompress to memory anymore?

Ah, sorry for being imprecise. Of course you still can decompress to normal RAM.
What I meant is: when decompressing to IO ($9fxx) it assumes you're decompressing to VRAM, not to some other IO device.
Honestly, since decompressing needs writing and reading back of already decompressed bytes, I don't see this working for anything else than VERA.
So when decompressing to VRAM, r1 should point to $9F23 and vera data0 to the address to uncompress to, with autoincrement 1. During decompression it will also use $9F24 for reading

kernal/lzsa.s
```diff
@@ -234,7 +213,7 @@ combinedbitz:
 	rts
 
 decompression_done:
-	PopW r3
+	stz VERA_CTRL			; could also push/pull, but this is shorter, KERNAL assumes D0..
```
irmen, line 216: This is what triggered my remark about it seemingly being only able to now copy into Vera vram. 
Why is this here? It looks like the change really does interfere with the normal operation when not copying to regular ram?

bsb-RickD: I see / understand.
So how to handle the "vera state"? when decompressing to VRAM, the code expects that D0 points to the destination in VRAM to decompress to.
After the decompression, the D0 address will point to the end of the decompression, just like R1 points to the end after decompression to normal RAM. I guess that's a good thing, because it passes information to the caller and is symmetric to the RAM decompression mode.
However, VERA_ctrl and the D1 address get clobbered / modified when decompressing to VRAM, 
Should I save/restore the ctrl and D1 state (only for the VRAM decompression mode?)

See, I first wanted to fit the routine into the R41 kernal, so I really squeezed for size, thats why I got rid of some push/pop code

bsb-RickD: So I pushed changes to save/restore D1 address and Vera control

bsb-RickD:
Any news on reviewing this pull request?
I frequently use decompression the VRAM, and I'm sure the X16 community will find this feature useful.

gc266 (approved):

irmen:
huh. This has been here for quite a while, I thought it was already merged long ago

indigodarkwolf:
Wrong repo. Not sure why someone "approved" the PR.

gc266 (approved):

irmen:
Blast. I notice now, this is the pre-community repo

for posterity:  the active repo is over here https://github.com/X16Community/x16-rom

PR #373, opened 2022-12-18
[DOS] Add support to return cwd as a pseudo directory listing with $=C.

### LINKED ISSUE #370: RFE: method for fetching the current working directory
Right now we don't have a way to know the name of the directory we're in.  I don't think anything in the fat32 driver keeps the state either other than the cluster.  It appears to keep cwd_cluster, and without adding any additional state, we could find the name of it by traversing up one level and iterating over the entries to match the cluster number to the start lba of the dirent.

It seems CPU-inefficient to iterate, so perhaps we should store the name of the dirent found by find_dir within fat32_chdir.

This takes an entire page of kvars RAM, but I think this is useful.

We could poll the cwd through a special DOS filename (like "`$=C`" or something)?

mooinglemur:
This adds enough bread crumbs for user applications to determine the full cwd path.

Please review and let me know if this is an acceptable approach to the problem.  It does expose some implementation details (the raw cluster number) but it adds minimal code and does not require any additional BSS to track state.  It also puts the onus on the user application to piece the cwd together.

To do this, the application would do the equivalent of `DOS"$=C:*=D`, make note of the cwd cluster (shown after the volume label).  Then do a `DOS"CD:..`, and then repeat `DOS"$=C:*=D`, making note of the cwd cluster and also parsing the contents of the directory to find the basename of the previous cwd.

This potentially closes https://github.com/commanderx16/x16-rom/issues/370

image


mooinglemur:
This is definitely barebones.  A couple of possible enhancements I can think of are:

- allow `$=CT` to show timestamps and clusters in the same listing
- provide a way to restore state of the cwd to a saved state.  This seems like a convenient way to restore state after walking up the tree to assemble the cwd.  This would require additional BSS.  Perhaps we provide a `save` and `restore` function via DOS, similar in concept to `pushd` in bash.

irmen:
observations:

- it will theoretically work.
- parsing the directory contents is pretty cumbersome, as it has always been
- the overall routine will be pretty complicated to link up all id's and such,
- but I am okay with this burden lying on the user program rather than the kernal itself because I think very very few user programs actually need to know something about the "current directory"

mooinglemur:
I decided to do this a different way.  I changed `$=C` show the name of the cwd directly. This, by itself, doesn't require any additional state in the fat32 driver.

I also added the ability to bookmark the cwd with the `C>` command, and to restore from the bookmark with the `C<` command.  This requires 4 additional bytes in the `fs` struct (of which there is one instance, `cur_volume`).  The reason for this feature is to allow user applications to walk the filesystem tree to construct a full cwd path and quickly restore their original cwd without having to keep state and work through the tree in DOS to get back there.

image

mooinglemur:
Marking as draft since I might be able to remove some complexity here by iterating the entire path upwards in one directory listing.

mooinglemur:
I think this is the final form.  DOS walks up the entire tree in one directory listing.  I'm marking it as ready for review again.

image

irmen:
Works like a charm! X16shell now only has to print the names in each line returned by the command, which is much easier, see it in action here:

image

mooinglemur:
Closing due to inclusion at X16Community fork

PR #374, opened 2022-12-26
Future additions of BASIC statements and functions without breaking tokenized representation

stefan-b-jakobsson:
This code is intended to make it possible to add future BASIC statements and functions without breaking the tokenized representation of BASIC programs.

Currently there are some 25 extended BASIC statements and 7 extended BASIC functions in the master branch. The interpreter treats statements and functions differently. A function returns a value, a statement does not. The extended statements and functions are tokenized with two bytes, first $CE and then the index of the statement/function with bit 7 set. If you want to add a new BASIC statement to the end of the list of statements, that will affect and increase all function indexes by one. A BASIC program in tokenized form saved to disk in an earlier ROM version may not work properly in a newer ROM version.

To handle this the tokenization may be changed, for example as follows:

- $CE still marks the start of an extended statement or function
- If bit 7 of the next byte is set, it's an extended statement. This is no change compared to the current code
- But if bit 7 of the next byte is clear, it's an extended function. I don't think the index value 0 can be used because that indicates the end of a BASIC line. Therefore, the function index is increased by one

Examples of new tokenization:
```
- $CE $80 = MON ; unchanged, the first extended statement
- $CE $98 = BANK ; unchanged, the last extended statement
- $CE $01 = VPEEK ; changed, the first extended function 
- $CE $07 = BIN$ ; changed, the last extended function
```
This PR will of coarse in itself break backward tokenization compatibility, but hopefully this is the last time we need do that.

### DIFF
--- a/basic/code10.s
+++ b/basic/code10.s
@@ -221,8 +221,7 @@ snerr9:	jmp snerr
 	tay
 	lda ptrfunc,y
 	sta jmper+1
-	iny
-	lda ptrfunc,y
+	lda ptrfunc+1,y
 	sta jmper+2
 	jsr jmper
 	jmp chknum
--- a/basic/code10.s
+++ b/basic/code10.s
@@ -211,18 +211,18 @@ isfun
 	bne nesct3
 
 	jsr chrget
-	sec
-	sbc #$80 + num_esc_statements
-	bcs :+
+	bne :+
 snerr9:	jmp snerr
-:	cmp #num_esc_functions
+:	dea
+	cmp #num_esc_functions
 	bcs snerr9
 
 	asl a
 	tay
-	lda stmdsp2+2*num_esc_statements,y
+	lda ptrfunc,y
 	sta jmper+1
-	lda stmdsp2+2*num_esc_statements+1,y
+	iny
+	lda ptrfunc,y
 	sta jmper+2
 	jsr jmper
 	jmp chknum
--- a/basic/code10.s
+++ b/basic/code10.s
@@ -213,7 +213,8 @@ isfun
 	jsr chrget
 	bne :+
 snerr9:	jmp snerr
-:	dea
+:	sec
+	sbc #$c0
 	cmp #num_esc_functions
 	bcs snerr9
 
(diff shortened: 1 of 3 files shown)


mobluse:
What was the tokenization before for the changed tokens? 

I think that prg-files with BASIC on the commanderx16.com web site with "Try it now" should be automatically upgraded from R38 if a later release with this PR is used. E.g. VPEEK is used a lot.

stefan-b-jakobsson:
Presently there’s a common list for new statements and functions. The interpreter refers to each one by its index in that list.

For implementation reasons, all statements are listed before the functions. As new statements are added, this will change the function indexes.

Taking VPEEK as an example:

In R38 it was tokenized as $CE $91.

In R39 three new statements were added, and then it became $CE $94.

In R41 more statements were added, and VPEEK was now $CE $98.

And finally in the current master yet another statement has been merged making VPEEK be $CE $99.

Backward tokenization compatibility is already broken. This PR just tries to stop that for future additions.

mist64:
Great, except that we need to take care of one more detail: Back then, I picked the prefix $CE and the start token of $80 for a reason: The C128 uses the prefix $CE and codes $02-$0A:

https://www.c64-wiki.de/wiki/Token

and the MEGA65 extends this to $02-$0E:

https://mega65.org/assets/pdf/MEGA65_BASIC_65_Referenzhandbuch.pdf

I didn't want to clash with their encoding, which is why I started with codes $80+.

Can we do $80+ and $C0+ instead (64 statements and 64 functions)? Or $40+ and $80+ (128 statements and 64 functions, leaving 50 for MEGA65 growth)?

stefan-b-jakobsson:
OK, I will change it to $80/$C0.

stefan-b-jakobsson:
I've committed code with changed encoding.
Statements: $CE + $80-$BF
Functions: $CE + $C0-$FF

stefan-b-jakobsson:
One thought.

In the unlikely event we need more than 64 new statements, we could add statements to the top of the list. If one statement is added to the top and referenced by index $7F, it doesn’t affect any of the subsequent indexes.

Issue #328, opened 2022-06-19
RFE: ROM bank dedicated to lookup tables, for VERA note frequencies and perhaps other purposes


mooinglemur:
I'm opening this issue mainly for discussion, as the idea has been casually mentioned in the Discord.

As it stands, VERA PSG takes a frequency value which linearly corresponds to Hz.  This requires a lookup table to translate from note-pitch values to frequency.  For the pure note values, this requires roughly a page of lookups.

However, if someone wants to do pitch bends/slides on VERA PSG and the YM2151 at the same time, the YM uses a logarithmic note/note-fraction based value, while VERA PSG uses a linear Hz value, requiring a large lookup table.

To store one octave's worth of 1/64 semitone step frequencies, matching the resolution of the YM2151, 768 values are required, or 1.5 KiB, which can be ROR/LSRed into lower octaves as needed.  If we wanted to cut the resolution to 1/16 semitone, we could use 192 values instead, or 1/20 of a semitone using 240 values.

I don't know:
* whether the powers that be are open to the idea of such a ROM bank
* the acceptable amount of duplication in such a bank to support saving CPU cycles for as many use cases as possible
* what everyone else would want in such a bank besides these freq lookups

Unfortunately, whatever is done here will likely need to be set in stone once it's in the bank.  Locations can't change without breaking code that uses of the values from the bank, so decisions for this would need careful consideration.

irmen:
Currently I have to use floating point math to convert 
https://github.com/irmen/prog8/blob/master/examples/cx16/bdmusic.p8#L34

I could create a lookup table myself as the notes for this particular example are all in a table already, but I'd like a general solution too

mooinglemur:
https://github.com/commanderx16/x16-rom/pull/376 closes this issue

PR #376, opened 2023-01-03
New audio bank and API

### LINKED ISSUE #330: reset doesn't silence the vera PSG

mooinglemur:
First of all, my apologies for the size of this PR.  This has been a fairly substantial effort and a lot more of this API came together more quickly than anticipated, and I believe it is feature complete with what @ZeroByteOrg @jestin and I have set out to do, and then some.

This closes https://github.com/commanderx16/x16-rom/issues/330 as this bank's init API is called from the KERNAL's reset handler.

Other than the init routine, we've tried to touch the other banks as little as possible.  We took 2 bytes of zeropage at the end of the reserved KERNAL area ($A7-$A8).  We reserved a page of Bank 0 RAM for shadowing writes to the YM2151 chip, and we've sliced off another 64 bytes for internal state for some of the API routines.

in the `audio/` subdirectory, we have two scripts that spit out ca65 code.  One (`create_luts.py`) is for creating the frequency lookup tables (`audio_luts.s`) used by some of the routines, and the other `fur2ymp.py` is for creating all of the instrument patches (`fmpatches.s`) from the one small binary data file in the `audio/` directory: `GM_X16.fur`.

`GM_X16.fur` is the data file containing the source of all of the instrument patches, and using [furnace](https://github.com/tildearrow/furnace/) is by far the easiest way to tweak and modify the instruments in the patch set. Thus I consider `GM_X16.fur` to be the closest thing to "source code" for this purpose, and I think it should be included in the repo for that reason.

The audio bank ($0A) has kernel API support included for user convenience, so machine language programs can switch to ROM bank $0A and still access the kernal jump table, similarly to how the BASIC bank can do the same.

Marking as a draft initially so that we can get more eyes for a bit before marking it as ready for review.

Following this PR, we expect to create another one with added BASIC keywords to make use of this new API.


ZeroByteOrg:
Additionally, there will be a PR for x16-docs which updates the audio programming chapter to document the audio bank API.
https://github.com/mooinglemur/x16-docs/blob/20230101-audio/X16%20Reference%20-%2009%20-%20Sound%20Programming.md

ZeroByteOrg:
Also, the intention is that once this new audio bank is approved and merged, that there is a PR to add several BASIC commands for audio, doing for sound what the graphics commands did for graphics. We felt that it made the most sense to keep the BASIC stuff separate from this PR.

irmen:
This PR is the one that supersedes my smaller PR #364 that was meant to fix #330 initially.
The initialization code from that older PR is all in here as well and silences the audio chips on reset.

mooinglemur:
This would close https://github.com/commanderx16/x16-rom/issues/328

mooinglemur:
closing due to inclusion in X16Community fork

PR #377, opened 2023-01-06
Enable kernal to accurately detect amount of memory installed

JimmyDansbo:
ROM was relying on the fact that the emulator wraps around to bank 0 when no more banks of memory are present. Real hardware does not have this functionality instead, writes to empty banks are not stored anywhere. For this reason I have added a bit of code that checks that a write to a bank can actually be read back, then it compares it with value on bank 0 to ensure continued support in the emulator as well.
I have tested the ROM in latest emulator and on unoffical hardware designed by Joe Burks (Wavicle)

### DIFF
--- a/kernal/drivers/x16/memory.s
+++ b/kernal/drivers/x16/memory.s
@@ -79,26 +79,54 @@ ramtas:
 ;
 ; detect number of RAM banks
 ;
-	stz ram_bank
-	ldx $a000
-	inx
-	lda #1
-:	sta ram_bank
-	ldy $a000
-	stx $a000
-	cpx $a000	; Ensure that value is actually written
-	bne :+		; Otherwise we have reached end of RAM
-	stz ram_bank
-	cpx $a000
-	sta ram_bank
-	sty $a000
+	ldx #$01	; 0K-512K
+	stx ram_bank
+	jsr :+
+	ldx #$40	; 512K-1024K
+	stx ram_bank
+	jsr :+
+	ldx #$80	; 1024K-1536K
+	stx ram_bank
+	jsr :+
+	ldx #$C0	; 1536K-2048K
+	stx ram_bank
+	jsr :+
+	lda #$00
+	jsr :+++++
+
+:	ldy $a000	; Save value so it can be restored later
+	lda $a000	; Load value, change with xor and store new value
+	eor #$ff
+	sta $a000
+	cmp $a000	; If values are not equal, memory is non-existing
+	bne :+
+	stz ram_bank	; Check for wrap-around
+	cmp $a000	; If values are equal, we have wrapped around
+	stx ram_bank
 	beq :+
-	inc		; check all banks instead of only mutiples of 8
-	bne :-
+	sty $a000	; Restore original value
+
+	ldy $a001	; Save value so it can be restored later
+	lda $a001	; Load value, change with xor and store new value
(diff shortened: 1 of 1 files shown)


mooinglemur:
I feel that this check could rely too much on reading a specific value out of RAM without checking that the RAM actually changed from its previous value.  Open bus behavior could result in a false positive for the availability of a bank, on average 1 out of 256 times, of course depending on the data bus's previous values.

It seems like we should do multiple write-reads of each bank's $a000 for more robustness

I think the original check had similar deficiencies, though, so this is nothing new.  But rather than showing, for instance, 1024K of RAM on a 512K system, it would show 520K.

mooinglemur:
Perhaps something like this would be more robust?  Just an idea :)
```
lda $a000
eor #$ff
sta $a000
cmp $a000
bne whatever
; put it back the way it was
eor #$ff
sta $a000
```

JimmyDansbo:
I think it may be a question of reliability vs. size, but I agree that relying on the, more or less, random value obtained by reading $a000 on bank0 and incrementing it is not optimal.

ZeroByteOrg:
Would it make sense to have some "sentinel value" in $00:A000 and then subsequent writes to other banks should be different than both the initial value AND the sentinel value?
I.e. read $00:A000 and then $01:A000. INC $01:A000, read $00:A000 (to "cleanse the palette of the bus) and then re-read $01:A000. If new value is returned, DEC $01:A000 and then move to next bank.
I'd imagine that banks could be checked as $00, $40, $80, $C0 and be safe to determine RAM amount, as the system uses 512K chips in 4 sockets.

JimmyDansbo:
I have been trying to keep the code as close to the original as possible without using more ROM space than absolutely necessary. I believe that my current entry adds 3 bytes to the ROM. I think the idea of "cleanse the pallette of the bus" is good, but I can't figure a way of doing it without adding quite a bit of code.
Also we need to keep in mind that the emulator will just wrap around to bank zero if there is no RAM at the tested bank, i.e. Startet without the -ram option, banks $40, $80 & $C0 wrap to bank $00. With -ram 1024, banks $80 & $C0 wrap to $00.
As the system is not supported with less than 512KB of memory, it would actually suffice to test banks $40, $80 and $C0 unless we want the kernal to be able to accurately detect the amount of memory available even if it is an "unsupport" amount.

indigodarkwolf:
If the emulator's behavior is different from the hardware here, we should change the emulator's behavior. This should be a simple change.

JimmyDansbo:
Wavicle tested on prototype3 and at the moment the kernal believes that it has 2048KB even though there is only 512KB. I believe that emulator should be changed to not expect any wrap-around at all and I will try to write some new detection code that uses the EOR function to test banks $00, $40, $80 & $C0

irmen:
@JimmyDansbo it may be easier for you if you're not making this PR on your master branch but instead use a feature branch to stick the changes on... ?   (now you can't mix it with other things that are happening on the master branch perhaps.)

JimmyDansbo:
@irmen I am not really familiar with how to do things on github, but I do see your point. In the future I will see if I can create a branch before creating a pull request. For the moment I don't really have anything else I can contribute to the ROM so lets just hope that this gets merged before I find something else I would like to work on.

JimmyDansbo:
Closing Pull request as it has been merged

PR #378, opened 2023-01-15
BASIC sound

jestin:
This is to add a set of BASIC statements for taking advantage of the proposed audio bank (https://github.com/commanderx16/x16-rom/pull/376).  These BASIC statements were the original purpose for the audio bank before it was realized that the audio bank could be even more useful as an extension to the kernal API and not just an extension to BASIC.

This pull request is dependent on the audio bank being merged, but beyond that has no other dependencies.  Upon merging of pull request [376](https://github.com/commanderx16/x16-rom/pull/376), this pull request will be rebased on top of the lastest in `master` so that it includes only the changes to BASIC.

The additions to BASIC in this pull request are as follows:

- `FMINIT` - Initialize the YM2151 and the audio bank shadow ram from BASIC (also useful for muting all channels in a single statement)
- `FMNOTE` - Play a single note on an FM channel.  This can be used to begin playing a note, change the pitch of an already playing note, or to stop playing a note
- `FMDRUM` - Trigger a percussion hit from the set of percussion instruments included in the audio bank.  This both sets the patch for a channel and plays the note (drum hit) in a single statement.
- `FMINST` - Assign a patch from the audio bank to an FM channel
- `FMVIB` - Set the speed and the depth of the LFO on the YM2151.  This LFO is used by patches configured for it, and several such patches are included in the audio bank.
- `FMFREQ` - Play a specific frequency in hertz on an FM channel.  This is analogue to `FMNOTE`, except the user can use hertz instead of musical notes.
- `FMVOL` - Set the volume of an FM channel
- `FMPAN` - Set whether a channel plays from the left speaker, the right speaker, neither, or both.
- `FMPLAY` - Plays a series of notes in a specified rhythm, defined by a string notation similar to that used by GWBASIC.
- `FMCHORD` - Plays multiple notes simultaneously on contiguous channels.  Similar to calling `FMNOTE` multiple times in  a row on different channels.
- `FMPOKE` - Write a byte directly to one of the YM2151's registers using the audio bank's shadow ram and wait loop write.  This allows full access to the YM2151 hardware while still be simple to call from BASIC.
- `PSGINIT` - Initialize the VERA's PSG (also useful for muting all channels in a single statement)
- `PSGNOTE` - Play a single note an a PSG channel.  This can be used to begin playing a note, change the pitch of an already playing note, or to stop a playing note.
- `PSGVOL` - Set the volume of a PSG channel.
- `PSGWAV` - Set the waveform and pulse width of a PSG channel.
- `PSGFREQ` - Play a specific frequency in hertz on a PSG channel.  This is an analogue to `PSGNOTE`, except the user can use hertz instead of musical notes.
- `PSGPAN` - Set whether a channel plays from the left speaker, the right speaker, neither, or both.
- `PSGPLAY` - Plays a series of notes in a specified rhythm, defined by a string notation similar to that used by GWBASIC.
- `PSGCHORD` - Plays multiple notes simultaneously on contiguous channels.  Similar to calling `PSGNOTE` multiple times in a row on different channels.

This PR was worked on in conjunction with the development of the audio bank, and was a collaboration between @mooinglemur, @ZeroByteOrg, and myself.  Please reach to me or my collaborators if you feel any modifications are required.


basic/basic.s
```diff
@@ -35,4 +35,5 @@
 .include "init.s"
 .include "x16additions.s"
 .include "graph.s"
+.include "sound.s"
```
jestin, line 38: Following the convention of the new X16 graphics statements, I created all new BASIC statements for sound in their own file.

basic/token2.s
```diff
@@ -82,6 +82,25 @@ reslst2	.byt "MO", 'N' + $80
...
+	.byt "FMCHOR", 'D' + $80
+	.byt "FMPOK", 'E' + $80
+	.byt "PSGINI", 'T' + $80
+	.byt "PSGNOT", 'E' + $80
+	.byt "PSGVO", 'L' + $80
+	.byt "PSGWA", 'V' + $80
+	.byt "PSGFRE", 'Q' + $80
+	.byt "PSGPA", 'N' + $80
+	.byt "PSGPLA", 'Y' + $80
+	.byt "PSGCHOR", 'D' + $80
```
jestin, line 103: I added all new statements after the existing statements, but before the existing functions.

basic/token2.s
```diff
@@ -187,6 +214,22 @@ brktxt	.byt $d
...
+err35	.byt "INVALID VOLUM", 'E'+$80
+ervol	=35
+err36	.byt "INVALID DRU", 'M'+$80
+erdrum	=36
+err37	.byt "INVALID DEPT", 'H'+$80
+erdep	=37
+err38	.byt "INVALID FREQUENC", 'Y'+$80
+erfrq	=38
+err39	.byt "INVALID PA", 'N'+$80
+erpan	=39
```
jestin, line 232: New error messages were defined in order to display friendly messages to users.  Up front validation is typically done prior to calling the routines from the audio bank, and that is where these messages are used.  If space in the BASIC bank becomes an issue, these friendly messages could be removed and the new BASIC statements could rely on the validation done in the audio bank (which typically sets the carry flag when called with bad input).  That would make the new BASIC statements less friendly, but would save space in the BASIC bank.

jestin:
Closed because this has now been merged to the github.com/X16Community/x16-rom fork of this repository.

